PR #64, opened 2020-01-17
Add colliders to the duck obj

kaikecarlos:


kaikecarlos:
Maybe, I didn't tried with sphere collider.

amazingalek:
The sphere collider was already there, you replaced it with a box collider :P

PR #84, opened 2020-03-13
Allow patching by providing MethodInfo

Raicuparta:
This way we can easily patch static types, or provide method arguments for more precise selection.

```C#
var getTexture = typeof(InputTranslator).GetMethod("GetButtonTexture", new[] { typeof(XboxButton) });
HarmonyHelper.AddPostfix(getTexture, typeof(Patches), nameof(PostGetTexture));
```

### DIFF
--- a/OWML.Common/IHarmonyHelper.cs
+++ b/OWML.Common/IHarmonyHelper.cs
@@ -1,12 +1,17 @@
 ﻿using System;
+using System.Reflection;
 
 namespace OWML.Common
 {
     public interface IHarmonyHelper
     {
         void AddPrefix<T>(string methodName, Type patchType, string patchMethodName);
+        void AddPrefix(MethodInfo methodInfo, Type patchType, string patchMethodName);
         void AddPostfix<T>(string methodName, Type patchType, string patchMethodName);
+        void AddPostfix(MethodInfo methodInfo, Type patchType, string patchMethodName);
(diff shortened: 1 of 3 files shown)


OWML.Common/IHarmonyHelper.cs
```diff
@@ -1,12 +1,17 @@
 ﻿using System;
+using System.Reflection;
 
 namespace OWML.Common
 {
     public interface IHarmonyHelper
     {
         void AddPrefix<T>(string methodName, Type patchType, string patchMethodName);
+        void AddPrefix(MethodInfo methodInfo, Type patchType, string patchMethodName);
```
amazingalek: How about this instead?
`void AddPrefix(Type type, string methodName, Type patchType, string patchMethodName);`
That is, allowing type as first parameter instead of the generic. Then calling it would be the more similar to the old/generic way.

amazingalek: It would also require fewer changes other places

Raicuparta: @amazingalek My point with this way is that it simultaneously takes care of static classes, *and* for when you need to pass parameters to select the method. If you want to have type as an argument instead, then you'll need 4 variations of each method: with and without type, with and without parameters. This way you just need two. Plus, this means you don't have to keep adding parameters if we eventually find we need other ways of finding the methods.

PR #92, opened 2020-04-27
Add mod load order and mod interaction methods

misternebula:


OWML.ModHelper/ModManifest.cs
```diff
@@ -24,6 +24,9 @@ public class ModManifest : IModManifest
         [JsonProperty("owmlVersion")]
         public string OWMLVersion { get; private set; }
 
+        [JsonProperty("dependencies")]
+        public string[] Dependencies { get; private set; }
```
amazingalek, line 28: if you remove "private" then you can set this from the outside and remove SetDependencies. Not a big deal, if you wanna keep this, then it's fine ;)

OWML.ModHelper.Interaction/ModInteraction.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        public bool ModExists(string uniqueName)
+        {
+            return _finder.GetMods().Where(m => m.Manifest.UniqueName == uniqueName).Count() != 0;
+        }
+
+        public T InvokeMethod<T>(string uniqueName, string methodName, params object[] parameters)
+        {
+            var mod = _modList.Where(m => m.ModHelper.Manifest.UniqueName == uniqueName).First();
```
amazingalek, line 50: this line is repeated, can be a private method

amazingalek: Also, list.Where(...).First() is the same as list.First(...)

OWML.ModHelper/ModManifest.cs
misternebula: yeah, just thought it'd be more in line with the setsettingsvalue in the config one

OWML.Common/packages.config
```diff
@@ -1,4 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="OW.Unity.Dlls" version="1.0.2" targetFramework="net35" />
+  <package id="Json.Net.Unity3D" version="9.0.1" targetFramework="net35" />
```
amazingalek, line 3: I wonder why this is needed now and not before. Just commenting this so I can check it again later ;)

OWML.ModHelper.Assets/packages.config
```diff
@@ -1,4 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="OW.Unity.Dlls" version="1.0.2" targetFramework="net35" />
+  <package id="Json.Net.Unity3D" version="9.0.1" targetFramework="net35" />
```
amazingalek, line 3: Yeah, this is now referenced in all projects. Can probably be removed

OWML.ModHelper.Interaction/ModInteraction.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        public ModInteraction(List<ModBehaviour> list, IModFinder finder)
+        {
+            _modList = list;
+            _finder = finder;
+
+            foreach (var mod in _finder.GetMods())
+            {
+                List<IModData> temp = new List<IModData>();
+                foreach (var mod2 in _finder.GetMods())
```
amazingalek, line 22: Double loop - can probably be simplified. Commenting so I can come back to it.

OWML.ModHelper.Interaction/ModInteraction.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        public IList<IModData> GetMods()
+        {
+            return _finder.GetMods();
+        }
+
+        public IList<IModData> GetDependants(string dependencyUniqueName)
+        {
+            return dependantDict.Where(x => x.Key.Manifest.UniqueName == dependencyUniqueName).FirstOrDefault().Value;
```
amazingalek, line 40: Can dependantDict.Where(...).FirstOrDefault() be null? If so this will crash

OWML.ModLoader/Owo.cs
```diff
@@ -38,8 +42,28 @@ public void LoadMods()
                 Application.logMessageReceived += OnLogMessageReceived;
             }
             var mods = _modFinder.GetMods();
-            foreach (var modData in mods)
+
+            var sortedMods = ModSorter.SortMods(mods);
```
amazingalek, line 46: I would prefer ModSorter to not be static. It can be newed up outside of Owo and injected in the constructor.

OWML.ModLoader/ModSorter.cs
```diff
@@ -0,0 +1,105 @@
...
+            };
+        }
+
+        private static IList<T> Sort<T, TKey>(IEnumerable<T> source, Func<T, IEnumerable<TKey>> getDependencies, Func<T, TKey> getKey)
+        {
+            return Sort<T>(source, RemapDependencies(source, getDependencies, getKey));
+        }
+    }
+
+    public class ModDep
```
amazingalek, line 91: ModDep can be in a separate file (ModDep.cs)

OWML.ModLoader/ModSorter.cs
```diff
@@ -0,0 +1,105 @@
...
+
+            foreach (var item in source)
+            {
+                Visit(item, getDependencies, sorted, visited);
+            }
+
+            return sorted;
+        }
+
+        private static void Visit<T>(T item, Func<T, IEnumerable<T>> getDependencies, List<T> sorted, Dictionary<T, bool> visited)
```
amazingalek, line 46: What does is "Visit" mean in this context?

misternebula: to look at an item

OWML.ModLoader/ModSorter.cs
```diff
@@ -0,0 +1,105 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace OWML.ModLoader
+{
+    class ModSorter
+    {
+        public static IList<ModDep> SortMods(IList<IModData> mods)
+        {
+            List<ModDep> modList = new List<ModDep>();
```
amazingalek, line 12: I prefer implicit types (var)

OWML.ModHelper.Interaction/ModInteraction.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        public bool ModExists(string uniqueName)
+        {
+            return _finder.GetMods().Where(m => m.Manifest.UniqueName == uniqueName).Count() != 0;
```
amazingalek, line 45: list.Where(...).Count() is the same as list.Count(...)

OWML.ModHelper.Interaction/ModInteraction.cs
```diff
@@ -0,0 +1,72 @@
...
+            return (T)mod.GetType().GetMethod(methodName).Invoke(mod, parameters);
+        }
+
+        public void InvokeMethod(string uniqueName, string methodName, params object[] parameters)
+        {
+            var mod = _modList.Where(m => m.ModHelper.Manifest.UniqueName == uniqueName).First();
+            mod.GetType().GetMethod(methodName).Invoke(mod, parameters);
+        }
+
+        public void SetVariableValue(string uniqueName, string variableName, object value)
```
amazingalek, line 60: How about calling this SetValue?

OWML.ModHelper.Interaction/ModInteraction.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            var mod = _modList.Where(m => m.ModHelper.Manifest.UniqueName == uniqueName).First();
+            mod.GetType().SetValue(variableName, value);
+        }
+
+        public T GetVariableValue<T>(string uniqueName, string variableName)
```
amazingalek, line 66: ... and GetValue?

OWML.ModHelper.Interaction/ModInteraction.cs
misternebula: It clashes with the existing setvalue from reflection I think, I'll have a look

OWML.ModHelper.Interaction/ModInteraction.cs
```diff
@@ -0,0 +1,72 @@
...
+        public T InvokeMethod<T>(string uniqueName, string methodName, params object[] parameters)
+        {
+            var mod = _modList.Where(m => m.ModHelper.Manifest.UniqueName == uniqueName).First();
+            return (T)mod.GetType().GetMethod(methodName).Invoke(mod, parameters);
+        }
+
+        public void InvokeMethod(string uniqueName, string methodName, params object[] parameters)
+        {
+            var mod = _modList.Where(m => m.ModHelper.Manifest.UniqueName == uniqueName).First();
+            mod.GetType().GetMethod(methodName).Invoke(mod, parameters);
```
amazingalek, line 57: Should handle cases where .GetMethod() is null. This goes for many of these methods.

OWML.Common/IModInteraction.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace OWML.Common
+{
+    public interface IModInteraction
```
amazingalek, line 8: Just a suggestion: instead of all this...
`var value = ModHelper.Interaction.GetVariableValue("foo", "bar");
ModHelper.Interaction.InvokeMethod("lolol");`

... how about this?
`var mod = ModHelper.Interaction.GetMod("foo");
var value = mod.GetVariableValue("bar");
mod.InvokeMethod("lolol");`

misternebula: so what would GetMod return? the modhelper?

OWML.Common/packages.config
amazingalek: I think this reference can be removed from most projects

OWML.ModLoader/ModSorter.cs
```diff
@@ -0,0 +1,105 @@
...
+                        Visit(dependency, getDependencies, sorted, visited);
+                    }
+                }
+
+                visited[item] = false;
+                sorted.Add(item);
+            }
+        }
+
+        private static Func<T, IEnumerable<T>> RemapDependencies<T, TKey>(IEnumerable<T> source, Func<T, IEnumerable<TKey>> getDependencies, Func<T, TKey> getKey)
```
amazingalek: Wow. Very cool. But a bit hard to read :P
Would it work if you for example called GetDependencies() first, then just gave the resulting list into here instead?

OWML.Common/IModInteraction.cs
```diff
@@ -0,0 +1,16 @@
...
+    {
+        IList<IModData> GetMods();
```
amazingalek, line 10: Would it be better to return IList<IModBehaviour> instead?

OWML.ModHelper.Events/TypeExtensions.cs
```diff
@@ -67,5 +67,11 @@ public static void Invoke(this object obj, string name, params object[] paramete
             method?.Invoke(obj, parameters);
         }
 
+        public static T Invoke<T>(this object obj, string name, params object[] parameters)
```
amazingalek, line 70: Nice 😁

OWML.ModHelper.Events/TypeExtensions.cs
```diff
@@ -67,5 +67,11 @@ public static void Invoke(this object obj, string name, params object[] paramete
...
+        {
+            var type = obj.GetType();
+            var method = type.GetAnyMethod(name);
+            return (T)method?.Invoke(obj, parameters);
```
amazingalek, line 74: I think this can be shortened by calling the other Invoke, then casting:
return (T)obj.Invoke(name, parameters);

misternebula: but that other invoke doesn't return anything

OWML.ModHelper.Interaction/ModInteraction.cs
```diff
@@ -0,0 +1,59 @@
...
+                        temp.Add(mod2);
+                    }
+                }
+                _dependantDict.Add(mod, temp);
+            }
+        }
+
+        public List<IModData> GetDependants(string dependencyUniqueName)
+        {
+            return _dependantDict.Where(x => x.Key.Manifest.UniqueName == dependencyUniqueName).FirstOrDefault().Value;
```
amazingalek, line 35: list.Where(...).FirstOrDefault() -> list.FirstOrDefault(...)

Also, what if none is found, doing .Value would crash, no?

OWML.ModHelper.Events/TypeExtensions.cs
misternebula: oh wait, i'm confused

OWML.ModHelper.Interaction/ModInteraction.cs
misternebula: hm, yeah, what does .FirstOrDefault() return on an empty list

OWML.ModHelper/ModHelper.cs
```diff
@@ -16,9 +16,10 @@ public class ModHelper : IModHelper
         public IModManifest Manifest { get; }
         public IModConfig Config { get; }
         public IOwmlConfig OwmlConfig { get; }
+        public IModInteraction ModInteraction { get; }
```
amazingalek, line 19: Maybe call it just Interaction here (and in the interface), so it's used like ModHelper.Interaction? Your call :)

OWML.ModHelper/ModManifest.cs
amazingalek: I see why you need to set it - in case it's null.
You can set the default value (in case of null) like this:
 public string[] Dependencies { get; private set; } = new[] { "None" }
Then you can remove SetDependencies and the place you call SetDependencies :)

OWML.ModLoader/ModSorter.cs
```diff
@@ -0,0 +1,90 @@
...
+    {
+        public IList<ModDep> SortMods(IList<IModData> mods)
+        {
+            var modList = new List<ModDep>();
+            foreach (var mod in mods)
+            {
+                if (mod.Manifest.Dependencies == null)
+                {
+                    mod.Manifest.SetDependencies(new[] { "None" });
+                }
```
amazingalek, line 18: This can be removed if you set default value, see above comment

OWML.ModLoader/ModSorter.cs
```diff
@@ -0,0 +1,90 @@
...
+
+                if (mod.Manifest.Dependencies.First() != "None")
+                {
+                    modList.Add(new ModDep(mod.Manifest.Name, mod, mod.Manifest.Dependencies));
+                }
+                else
+                {
+                    modList.Add(new ModDep(mod.Manifest.Name, mod));
+                }
```
amazingalek, line 27: 🤔 if Dependencies didn't default to { "none" }, couldn't you collapse these into just `modList.Add(new ModDep(mod.Manifest.Name, mod));` ?

OWML.ModLoader/Owo.cs
```diff
@@ -38,8 +44,28 @@ public void LoadMods()
...
+            var sortedMods = _sorter.SortMods(mods);
+
+            var modNames = new List<string>();
+            foreach (var mod in mods)
+            {
+                if (mod.Config.Enabled == true)
+                {
+                    modNames.Add(mod.Manifest.Name);
+                }
+            }
```
amazingalek, line 57: var modNames = mods.Where(mod => mod.Config.Enabled).Select(mod => mod.Manifest.Name);

OWML.ModLoader/Owo.cs
```diff
@@ -38,8 +44,28 @@ public void LoadMods()
...
+
+            foreach (var modDep in sortedMods)
             {
+                foreach (var dep in modDep.Dependencies)
+                {
+                    if (!modNames.Contains(dep) && dep != "None")
+                    {
+                        _console.WriteLine("Error! " + modDep.Name + " needs " + dep + ",  but it's disabled!");
```
amazingalek, line 65: You can use string interpolation:
_console.WriteLine($"Error! {modDep.Name} needs {dep},  but it's disabled!");

OWML.ModLoader/Owo.cs
```diff
@@ -38,8 +44,28 @@ public void LoadMods()
...
+                {
+                    modNames.Add(mod.Manifest.Name);
+                }
+            }
+
+            foreach (var modDep in sortedMods)
             {
+                foreach (var dep in modDep.Dependencies)
+                {
+                    if (!modNames.Contains(dep) && dep != "None")
```
amazingalek, line 63: Could you use null instead of "None" or does that have other consequences?

OWML.Common/IModManifest.cs
```diff
@@ -13,5 +13,8 @@ public interface IModManifest
         string UniqueName { get; }
         string ModFolderPath { get; set; }
         Dictionary<string, string> AppIds { get; }
+        string[] Dependencies { get; }
+
+        void SetDependencies(params string[] dependencies);
```
amazingalek, line 18: See comments on ModManifest.cs

OWML.ModHelper.Interaction/ModInteraction.cs
```diff
@@ -0,0 +1,72 @@
...
+        private readonly List<ModBehaviour> _modList = new List<ModBehaviour>();
+        private readonly IModFinder _finder;
+        private readonly Dictionary<IModData, List<IModData>> dependantDict = new Dictionary<IModData, List<IModData>>();
+
+        public ModInteraction(List<ModBehaviour> list, IModFinder finder)
+        {
+            _modList = list;
+            _finder = finder;
+
+            foreach (var mod in _finder.GetMods())
```
amazingalek, line 19: _finder.GetMods() is called many times but it's result doesn't change (right?). Can be cached to a variable first, which is reused.

OWML.ModHelper.Interaction/ModInteraction.cs
amazingalek: Nah, double loop is fine.

But variable names can be improved. Suggestion:

- temp -> dependants
- mod2 -> otherMod

OWML.ModHelper.Events/TypeExtensions.cs
amazingalek: You're right! Then you can do the opposite, call your new method from the original one (without a return) :)

amazingalek: Scratch that, let's not mess with existing functionality, just in case.

OWML.ModHelper.Interaction/ModInteraction.cs
amazingalek: Usually null

OWML.ModLoader/ModSorter.cs
misternebula: ModDep needs the dependencies to sort the mods, but if it's "None" then it doesn't add any dependencies

OWML.ModLoader/ModSorter.cs
```diff
@@ -0,0 +1,90 @@
...
+                        Visit(dependency, getDependencies, sorted, visited);
+                    }
+                }
+
+                visited[item] = false;
+                sorted.Add(item);
+            }
+        }
+
+        private Func<T, IEnumerable<T>> RemapDependencies<T, TKey>(IEnumerable<T> source, Func<T, IEnumerable<TKey>> getDependencies, Func<T, TKey> getKey)
```
amazingalek, line 75: This is some rocket science stuff. Very cool. Not very easy to read, though :P

Would it still work if you for example called GetDependencies before calling this method, then passed in the result instead? Just an example.

misternebula:
- Tested mod load order and dependency checking.

misternebula:
- Tested GetMod(), GetMods(), GetDependants(), GetDependencies() and Invoke<T>().

OWML.ModHelper.Interaction/ModInteraction.cs
misternebula: Nope. dependantDict contains every mod, so it'll only be empty if there are no mods. And if there's no mods, there's nothing to call this :P

OWML.ModHelper.Interaction/ModInteraction.cs
misternebula: Will just return an empty list if there are no dependants.

OWML.ModHelper/ModManifest.cs
```diff
@@ -27,6 +27,9 @@ public class ModManifest : IModManifest
         [JsonProperty("dependencies")]
         public string[] Dependencies { get; private set; } = new string[] { };
 
+        [JsonProperty("priorityLoad")]
+        public bool PriorityLoad { get; private set; } = false;
```
amazingalek, line 31: false is default for booleans, so you can skip this (or keep it if you want)

OWML.ModHelper.Interaction/ModInteraction.cs
amazingalek: If that's the case then you can use First instead of FirstOrDefault ;)

OWML.ModHelper.Interaction/ModInteraction.cs
```diff
@@ -1,36 +1,58 @@
...
 using System.Linq;
+using UnityEngine;
 
 namespace OWML.ModHelper.Interaction
 {
     public class ModInteraction : IModInteraction
     {
-        private readonly List<ModBehaviour> _modList = new List<ModBehaviour>();
-        private readonly Dictionary<IModBehaviour, List<IModBehaviour>> _dependantDict = new Dictionary<IModBehaviour, List<IModBehaviour>>();
+        private readonly IList<ModBehaviour> _modList;
```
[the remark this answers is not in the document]
amazingalek: Same goes for most places, try to use interfaces if possible

OWML.ModHelper.Interaction/ModInteraction.cs
```diff
@@ -1,36 +1,58 @@
...
             foreach (var mod in _modList)
             {
-                var dependants = new List<IModBehaviour>();
+                var dependants = new List<ModBehaviour>();
+                var dependencies = new List<ModBehaviour>();
                 foreach (var dependency in _modList)
-                {
-                    if (dependency.ModHelper.Manifest.Dependencies != new string[] { } && dependency.ModHelper.Manifest.Dependencies.Contains(mod.ModHelper.Manifest.Name))
+                { 
+                    if (dependency.ModHelper.Manifest.Dependencies != new string[] { } && dependency.ModHelper.Manifest.Dependencies.Contains(mod.ModHelper.Manifest.UniqueName))
```
amazingalek, line 31: `dependency.ModHelper.Manifest.Dependencies != new string[] { }` can be simplified to `dependency.ModHelper.Manifest.Dependencies.Any()` (.Any() is same as .Count == 0)

OWML.ModHelper.Interaction/ModInteraction.cs
```diff
@@ -1,36 +1,58 @@
...
                 }
                 _dependantDict.Add(mod, dependants);
+                _dependencyDict.Add(mod, dependencies);
             }
         }
 
         public IList<IModBehaviour> GetDependants(string dependencyUniqueName)
         {
-            return _dependantDict.Where(x => x.Key.ModHelper.Manifest.UniqueName == dependencyUniqueName).FirstOrDefault().Value;
+            RegenDicts();
```
amazingalek, line 48: Why do you have to regenerate the dictionaries, can they change? The constructor will run before anyone can call GetDependants/GetDependencies

misternebula: because of how mod loading works in general, the ModInteraction is made before the modbehaviour. that means that in the constructor, modinteraction only has access to all the mods that loaded *before* it loaded. this should be the same for _modList, but somehow it updates?? but the dict doesn't

OWML.ModHelper.Interaction/ModInteraction.cs
misternebula: shouldn't it be !dependency.ModHelper.Manifest.Dependencies.Any() ?
i'm reading that .Any() returns true if there are any elements in a sequence

misternebula: oh wait, nevermind

misternebula: done!

OWML.ModHelper.Interaction/ModInteraction.cs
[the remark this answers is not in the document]
misternebula: done

OWML.ModLoader/ModSorter.cs
```diff
@@ -0,0 +1,78 @@
...
+                if (inProcess)
+                {
+                    throw new ArgumentException("Cyclic dependency found.");
+                }
+            }
+            else
+            {
+                visited[item] = true;
+
+                var dependencies = getDependencies(item);
```
Raicuparta, line 49: Trying to understand why you can't just do `var dependencies = item.Dependencies` here and get rid of this getDependencies argument completely. Same with getKey argument in other places. Is there some place I'm not seeing where you pass a different way of getting these attributes? These arguments are making this file very hard to follow and I don't see why.

misternebula: believe this is fixed in the rewrite

OWML.ModLoader/ModSorter.cs
misternebula: this sorter has been rewritten now

OWML.ModLoader/ModSorter.cs
misternebula: this sorter has been rewritten now

OWML.ModLoader/ModSorter.cs
```diff
@@ -0,0 +1,122 @@
...
+            }
+            else
+            {
+                // return sortedList (a topologically sorted order)
+                return sortedList;
+            }
+        }
+    }
+
+    public class Tuple<T1, T2>
```
Raicuparta, line 103: Do we only use tuples for graph edges? If so, why not call this class `Edge`?
And, if tuples are only used for edges, they also only need to work with nodes (which are actually just strings?). So they don't need to be generic. You can just define the class Edge as having attributes `string From` and `string To`, instead of the more generic `Item1` and `Item2`.

In other situations it may make sense to have something super generic to work with everything, but in our case we know that we only want to use this sort for sorting mods, so we can type and name everything accordingly. This will make the code a lot more readable and reduce the type parameter noise.

OWML.ModLoader/Owo.cs
```diff
@@ -37,19 +43,36 @@ public void LoadMods()
...
+            var sortedPriority = _sorter.SortMods(priorityMods);
+
+            var modNames = _modFinder.GetMods().Where(mod => mod.Config.Enabled).Select(mod => mod.Manifest.UniqueName);
+
+            var sortedMods = Enumerable.Concat(sortedPriority, sortedNormal);
+
+            foreach (var mod in sortedMods)
             {
-                var modType = LoadMod(modData);
+                foreach (var dep in mod.Manifest.Dependencies)
```
Raicuparta, line 58:
```suggestion
                foreach (var dependency in mod.Manifest.Dependencies)
```
Unnecessary abbreviations make code less readable below

PR #93, opened 2020-04-29
Mod interaction

### LINKED ISSUE #79: Mod dependencies
* Allow mod dependencies to be defined in manifest
* Load mods in order
* Helper methods to interact with other mods

misternebula:
Closes #79


OWML.ModHelper/ModManifest.cs
```diff
@@ -24,6 +24,12 @@ public class ModManifest : IModManifest
         [JsonProperty("owmlVersion")]
         public string OWMLVersion { get; private set; }
 
+        [JsonProperty("dependencies")]
+        public string[] Dependencies { get; private set; } = new string[] { };
+
+        [JsonProperty("priorityLoad")]
+        public bool PriorityLoad { get; private set; } = false;
```
amazingalek, line 31: bools default to false, so remove `= false;` if you want

OWML.ModLoader/ModDep.cs
```diff
@@ -0,0 +1,23 @@
+﻿using OWML.Common;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace OWML.ModLoader
+{
+    public class Mod
```
amazingalek, line 9: Like (I think) @Raicuparta suggested, you can change both the filename and class name to ModDependency

OWML.ModLoader/ModSorter.cs
```diff
@@ -0,0 +1,122 @@
+﻿using OWML.Common;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace OWML.ModLoader
+{
+    class ModSorter
```
amazingalek, line 8: public/private?

OWML.ModLoader/ModSorter.cs
```diff
@@ -0,0 +1,122 @@
...
+
+            // Make list of uniqueNames
+            var modList = new List<string>();
+            foreach (var mod in mods)
+            {
+                 modList.Add(mod.Manifest.UniqueName);
+            }
+
+            // Make hashset of tuples of Dependant : Dependency
+            HashSet<Tuple<string, string>> set = new HashSet<Tuple<string, string>>();
```
amazingalek, line 27: var

OWML.ModLoader/ModDep.cs
misternebula: i... did? probably didn't save when I switched branches...

misternebula: oh wait, this was before the commit

OWML.ModLoader/ModSorter.cs
```diff
@@ -0,0 +1,121 @@
...
+            }
+            else
+            {
+                // return sortedList (a topologically sorted order)
+                return sortedList;
+            }
+        }
+    }
+
+    public class Edge<T1, T2>
```
Raicuparta, line 102: Now that `First` and `Second` are just strings, we don't need this class to have the type parameters  `<T1, T2>`, you can remove them here. This means that when you create an edge, you can just do `new Edge(first, second)`.

OWML.ModHelper.Interaction/ModInteraction.cs
```diff
@@ -0,0 +1,79 @@
...
+                }
+                _dependantDict.Add(mod, dependants);
+                _dependencyDict.Add(mod, dependencies);
+            }
+        }
+
+        public IList<IModBehaviour> GetDependants(string dependencyUniqueName)
+        {
+            RegenDicts();
+            return _dependantDict.Where(x => x.Key.ModHelper.Manifest.UniqueName == dependencyUniqueName).FirstOrDefault().Value;
```
Raicuparta, line 49: Seems like we're using dictionaries to store the dependencies, but then we're not taking advantage of the fact that they are dictionaries. I think we can change the dependency and dependant dictionary definition to `Dictionary<string, List<IModBehaviour>>`, where the string is the uniqueName. So when you need to get a dependency / dependant you do `_dependantDict[dependencyUniqueName]`.

OWML.ModLoader/ModSorter.cs
```diff
@@ -0,0 +1,121 @@
...
+            // Get the IModData back by looking up uniqueName in dict
+            var returnList = new List<IModData>();
+            sortedList.ForEach(mod => returnList.Add(modDict[mod]));
+
+            return returnList;
+        }
+
+        // Thanks to https://gist.github.com/Sup3rc4l1fr4g1l1571c3xp14l1d0c10u5/3341dba6a53d7171fe3397d13d00ee3f
+
+        static List<string> TopologicalSort<T>(HashSet<string> nodes, HashSet<Edge<string, string>> edges)
```
Raicuparta, line 53: Type parameter `T` no longer needed (since all nodes are strings)

OWML.ModHelper.Interaction/ModInteraction.cs
```diff
@@ -0,0 +1,79 @@
...
+                    }
+                }
+                _dependantDict.Add(mod.ModHelper.Manifest.UniqueName, dependants);
+                _dependencyDict.Add(mod.ModHelper.Manifest.UniqueName, dependencies);
+            }
+        }
+
+        public IList<IModBehaviour> GetDependants(string dependencyUniqueName)
+        {
+            RegenDicts();
```
amazingalek, line 48: Does RegenDicts need to be called every time, or could you skip it if the dictionaries already have data?

OWML.ModHelper.Interaction/ModInteraction.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        public IList<IModBehaviour> GetMods()
+        {
+            return _modList;
+        }
+
+        public bool ModExists(string uniqueName)
+        {
+            return _modList.Count(m => m.ModHelper.Manifest.UniqueName == uniqueName) != 0;
```
amazingalek, line 76: `list.Count(...) != 0` -> `list.Any(...)`

OWML.ModLoader/ModSorter.cs
```diff
@@ -0,0 +1,132 @@
...
+            sortedList.ForEach(mod => returnList.Add(modDict[mod]));
+
+            return returnList;
+        }
+
+        // Thanks to https://gist.github.com/Sup3rc4l1fr4g1l1571c3xp14l1d0c10u5/3341dba6a53d7171fe3397d13d00ee3f
+
+        static List<string> TopologicalSort(HashSet<string> nodes, HashSet<Edge> edges)
+        {
+            // Empty list that will contain the sorted elements
```
amazingalek, line 75: You can remove all these comments. The could *should* be self explanatory ;)

Readme.md
```diff
@@ -69,6 +69,7 @@ Each mod is defined in a manifest.json file:
 |uniqueName|Usually {author}.{uniqueName}.|
 |version|The version number.|
 |owmlVersion|The version of OWML the mod was built for.|
+|dependencies|\Array of dependency names. Make sure to use the unique name.|
```
amazingalek, line 72: Remove the `\`

OWML.ModLoader/ModSorter.cs
misternebula: Is it bad though? Surely if this code needs to be changed in the future, it's better the more readable it is?

OWML.ModHelper.Interaction/ModInteraction.cs
```diff
@@ -0,0 +1,82 @@
...
+        {
+            _dependantDict = new Dictionary<string, List<IModBehaviour>>();
+            _dependencyDict = new Dictionary<string, List<IModBehaviour>>();
+            foreach (var mod in _modList)
+            {
+                var dependants = new List<IModBehaviour>();
+                var dependencies = new List<IModBehaviour>();
+                foreach (var dependency in _modList)
+                { 
+                    if (dependency.ModHelper.Manifest.Dependencies.Any() && dependency.ModHelper.Manifest.Dependencies.Contains(mod.ModHelper.Manifest.UniqueName))
```
amazingalek, line 31: Pretty sure you can remove `dependency.ModHelper.Manifest.Dependencies.Any() && ` and it'll work the same.

OWML.ModHelper.Interaction/ModInteraction.cs
```diff
@@ -0,0 +1,82 @@
...
+                        dependants.Add(dependency);
+                    }
+
+                    if (mod.ModHelper.Manifest.Dependencies.Contains(dependency.ModHelper.Manifest.UniqueName))
+                    {
+                        dependencies.Add(dependency);
+                    }
+                }
+                _dependantDict.Add(mod.ModHelper.Manifest.UniqueName, dependants);
+                _dependencyDict.Add(mod.ModHelper.Manifest.UniqueName, dependencies);
```
amazingalek, line 42: doing .Add on a dictionary will fail if it already contains the key. If you want this to work with duplicate mods you can do `dict[key] = value` instead.

OWML.ModHelper.Interaction/ModInteraction.cs
```diff
@@ -0,0 +1,82 @@
...
+
+namespace OWML.ModHelper.Interaction
+{
+    public class ModInteraction : IModInteraction
+    {
+        private readonly IList<IModBehaviour> _modList;
+        private Dictionary<string, List<IModBehaviour>> _dependantDict = new Dictionary<string, List<IModBehaviour>>();
+        private Dictionary<string, List<IModBehaviour>> _dependencyDict = new Dictionary<string, List<IModBehaviour>>();
+
+        public ModInteraction(IList<IModBehaviour> list)
```
amazingalek, line 14: you could rename list and _modList to mods and _mods.

OWML.ModHelper.Interaction/ModInteraction.cs
```diff
@@ -0,0 +1,82 @@
...
+        {
+            _modList = list;
+
+            RegenDicts();
+        }
+
+        private void RegenDicts()
+        {
+            _dependantDict = new Dictionary<string, List<IModBehaviour>>();
```
amazingalek, line 23: you could rename _dependantDict and _dependencyDict  to _modDepandants and _modDependencies.

OWML.ModLoader/ModSorter.cs
```diff
@@ -0,0 +1,117 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace OWML.ModLoader
+{
+    public class ModSorter
+    {
+        private readonly IModConsole _console;
+        public ModSorter(IModConsole console)
```
amazingalek, line 11: would be nice with a line between these

OWML.ModLoader/ModSorter.cs
```diff
@@ -0,0 +1,117 @@
...
+                _console.WriteLine("Error - Cyclic dependency found. Returning original load order...");
+                return mods;
+            }
+
+            sortedList.Reverse();
+
+            var returnList = new List<IModData>();
+            sortedList.ForEach(mod => returnList.Add(modDict[mod]));
+
+            return returnList;
```
amazingalek, line 64: return sortedList.Select(mod => modDict[mod]);

amazingalek: you might need .ToList() at the end.

OWML.ModHelper.Interaction/ModInteraction.cs
Raicuparta: Good catch. We will eventually handle duplicate mods in the manager, but we shouldn't let OWML crash in this case.

amazingalek:
Awesome, all that's left is testing. :)
Unit tests would be nice but it's not a requirement 🤷

PR #108, opened 2020-06-03
Method summaries

misternebula:
- Still need to work on OWML.Common.Menus
- Haven't done return tags... most of the time it's really obvious though?

### DIFF
--- a/OWML.Common/Events.cs
+++ b/OWML.Common/Events.cs
@@ -5,28 +5,28 @@
     /// </summary>
     public enum Events
     {
-        /// <summary>Called before the Awake() method.</summary>
+        /// <summary>Called before MonoBehaviour.Awake().</summary>
         BeforeAwake = 0,
 
-        /// <summary>Called after the Awake() method.</summary>
+        /// <summary>Called after MonoBehaviour.Awake().</summary>
         AfterAwake = 1,
 
-        /// <summary>Called before the Start() method.</summary>
+        /// <summary>Called before MonoBehaviour.Start().</summary>
         BeforeStart = 2,
 
-        /// <summary>Called after the Start() method.</summary>
+        /// <summary>Called after MonoBehaviour.Start().</summary>
         AfterStart = 3,
 
-        /// <summary>Called before the OnEnable() method.</summary>
+        /// <summary>Called before MonoBehaviour.OnEnable().</summary>
         BeforeEnable = 4,
 
-        /// <summary>Called after the OnEnable() method.</summary>
+        /// <summary>Called after MonoBehaviour.OnEnable().</summary>
         AfterEnable = 5,
 
-        /// <summary>Called before the OnDisable() method.</summary>
+        /// <summary>Called before MonoBehaviour.OnDisable().</summary>
         BeforeDisable = 6,
 
-        /// <summary>Called after the OnDisable() method.</summary>
+        /// <summary>Called after MonoBehaviour.OnDisable().</summary>
         AfterDisable = 7
     }
 }
--- a/OWML.Common/Events.cs
+++ b/OWML.Common/Events.cs
@@ -1,14 +1,32 @@
 ﻿namespace OWML.Common
 {
+    /// <summary>
+    /// The types of events that can be subscribed to.
+    /// </summary>
     public enum Events
     {
+        /// <summary>Called before the Awake() method.</summary>
         BeforeAwake = 0,
+
+        /// <summary>Called after the Awake() method.</summary>
         AfterAwake = 1,
+
+        /// <summary>Called before the Start() method.</summary>
         BeforeStart = 2,
+
+        /// <summary>Called after the Start() method.</summary>
         AfterStart = 3,
+
+        /// <summary>Called before the OnEnable() method.</summary>
         BeforeEnable = 4,
+
+        /// <summary>Called after the OnEnable() method.</summary>
         AfterEnable = 5,
+
+        /// <summary>Called before the OnDisable() method.</summary>
         BeforeDisable = 6,
+
+        /// <summary>Called after the OnDisable() method.</summary>
         AfterDisable = 7
     }
 }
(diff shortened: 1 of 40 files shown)


OWML.Common/IHarmonyHelper.cs
```diff
@@ -3,15 +3,57 @@
 
 namespace OWML.Common
 {
+    /// <summary>
+    /// Handler for patching methods.
```
Raicuparta, line 7: Handler -> Helper?

OWML.Common/Events.cs
```diff
@@ -1,14 +1,32 @@
 ﻿namespace OWML.Common
 {
+    /// <summary>
+    /// The types of events that can be subscribed to.
+    /// </summary>
     public enum Events
     {
+        /// <summary>Called before the Awake() method.</summary>
```
Raicuparta, line 8: Maybe personal preference, but these could be simplified as just "Called before Awake()". Might also make sense to clarify that this is referring to the MonoBehaviour methods. So something like "Called before MonoBehaviour.Awake()". Leave it up to you to decide what you think makes most sense.

OWML.Common/IHarmonyHelper.cs
```diff
@@ -3,15 +3,57 @@
...
+    /// </summary>
     public interface IHarmonyHelper
     {
+        /// <summary>Add prefix patch to the given method.</summary>
+        /// <typeparam name="T">The type the method is in.</typeparam>
+        /// <param name="methodName">The name of the method.</param>
+        /// <param name="patchType">The type the patch is in.</param>
+        /// <param name="patchMethodName">The name of the patch method.</param>
```
Raicuparta, line 15: I think we can make the descriptions here a lot more helpful. Right now it does't seem to say much more than the parameter names already say. Not sure if it would be too much information to have here though. For instance:

1. The generic version of `AddPrefix` only works if there is only one method with the provided `methodName`;
2. The generic version also doesn't work if the provided type parameter `T` belongs to a static class;
3. If 1 or 2 are the case, we should use the non-generic version of AddPrefix, with a MethodInfo instead;
4. `patchMethodName` needs to be the name of a method in the `patchType` class;
5. The patch method needs to be static.

I think this would be enough information to make it possible to write a patch without having to read any documentation, except perhaps for knowing that HarmonyHelper.AddPrefix exists. Do you think we should keep this info for external documentation? I think it would be pretty cool to be able to understand all these restrictions right away, but might be overkill.

OWML.Common/IHarmonyHelper.cs
misternebula: Done.

OWML.Common/Events.cs
misternebula: Done.

OWML.Common/IHarmonyHelper.cs
misternebula: Done.

OWML.Common/IModAssets.cs
```diff
@@ -2,12 +2,35 @@
...
     {
+        /// <summary>Loads an asset bundle.</summary>
+        /// <param name="filename">The name of the asset bundle.</param>
+        /// <returns>The asset bundle.</returns>
         AssetBundle LoadBundle(string filename);
+
+        /// <summary>Loads a GameObject.</summary>
+        /// <param name="objectFilename">The name of the asset.</param>
+        /// <param name="imageFilename"></param>
+        /// <returns>The GameObject.</returns>
```
amazingalek, line 18: An empty game object containing the mesh and texture

OWML.Common/IModBehaviour.cs
```diff
@@ -2,11 +2,26 @@
...
 namespace OWML.Common
 {
+    /// <summary>The mod behaviour.</summary>
     public interface IModBehaviour
     {
+        /// <summary>The ModHelper attached to this mod.</summary>
         IModHelper ModHelper { get; }
+
+        /// <summary>The API attached to this mod.</summary>
+        object Api { get; }
```
amazingalek, line 12: So this PR builds upon your other mod interaction PR?

OWML.Common/IModConfig.cs
```diff
@@ -3,14 +3,36 @@
...
+    /// </summary>
     public interface IModConfig
     {
+        /// <summary>Is the mod enabled?</summary>
         bool Enabled { get; set; }
+
+        /// <summary>Does the mod use VR?</summary>
         bool RequireVR { get; set; }
+
+        /// <summary>The dict of key : value of settings. Use "GetSettingsValue" instead. </summary>
```
amazingalek, line 17: "Use "GetSettingsValue" instead." ? :)

OWML.Common/IModConfig.cs
```diff
@@ -3,14 +3,36 @@
...
         Dictionary<string, object> Settings { get; set; }
+
+        /// <summary>Get a value from the config file.</summary>
+        /// <typeparam name="T">The type of the value.</typeparam>
+        /// <param name="key">The name of the setting.</param>
+        /// <returns>The value of the setting.</returns>
         T GetSettingsValue<T>(string key);
+
+        /// <summary>Set a value in the config file.</summary>
```
amazingalek, line 26: Don't remember if this also updates/saves the file, are you sure?

OWML.Common/IModBehaviour.cs
misternebula: Yeah, the other PR would have to be merged first before this one

OWML.Common/Menus/IModButton.cs
```diff
@@ -1,6 +1,8 @@
 ﻿using System;
 using UnityEngine.UI;
 
+#pragma warning disable 1591
```
amazingalek, line 4: Disabling warnings? Not a fan ;)

amazingalek: The warnings are almost always correct, and reminds us to clean up our crap ^_^
If we ignore them, we'll never. Now? Maybe :P

amazingalek:
After merging with master, this'll shrink a bit :)

PR #112, opened 2020-06-22
Clearer errors on patch original method not found

Raicuparta:


amazingalek:
this is making things less clear for me :P

Raicuparta:
@amazingalek I agree that the code is more confusing now. The problem is that GetAnyMethod doesn't always throw an error when a method is not found, some times it returns null. So the try / catch and the null check both serve as a way to check if the method was found. I'm not sure how to improve it...

PR #113, opened 2020-06-22
fixed menus reacting to held button (switched from IsPressed to IsNewlyPressed)

TAImatem:
really, alek? who even uses IsPressed for buttons


amazingalek:
👍can you bump the version?

PR #124, opened 2020-06-27
Launch game with OpenVR flag if VR is required

### LINKED ISSUE #85: Run game with OpenVR flag if VR is required
In some cases, we need pass the parameter `-vrmode openvr` to the game exe in order for SteamVR to pick up the game. This can be done by passing the same param to OWML, but to make it simpler we should just always pass it whenever VR is required.

Raicuparta:
Closes #85


OWML.Launcher/App.cs
```diff
@@ -159,12 +167,17 @@ private void PatchGame(IList<IModData> mods)
...
+        private void StartGame(string[] args, bool enableVR)
         {
             _writer.WriteLine("Starting game...");
             try
             {
-                Process.Start($"{_owmlConfig.GamePath}/OuterWilds.exe", string.Join(" ", args));
+                var gameArgs = string.Join(" ", args);
+                if (enableVR)
+                {
+                    gameArgs += VrArgument;
```
amazingalek, line 178: shouldn't there be a space in front in some cases? 
if so, how about adding the arg to the list before joining?

Raicuparta: It's always the last argument, so no? I can do it via the array anyway. I started off that way but it looked messier to add stuff to the array than to the string 😅

amazingalek: ah, the vr arg has a space in front, so it doesn't matter 🤷

PR #134, opened 2020-06-28
Asset object fix

amazingalek:
this seems to fix all errors in verbose mode 🥳


OWML.SampleMods/OWML.LoadCustomAssets/LoadCustomAssets.cs
```diff
@@ -68,6 +68,7 @@ private void OnDuckLoaded(GameObject duck)
             duck.AddComponent<SphereCollider>();
             duck.AddComponent<Rigidbody>();
             _duckBody = duck.AddComponent<OWRigidbody>();
+            duck.SetActive(false);
```
Raicuparta, line 71: should this `SetActive` not come before adding OWRigidbody? I assume it works since you tested I thought OWRigidbody did some stuff on awake that we need to prevent.

amazingalek: Awake is run next frame (except we're deactivating it before that). Tested :)

PR #142, opened 2020-07-02
Revert "Launch game with OpenVR flag if VR is required"

Raicuparta:
Reverts amazingalek/owml#124


OWML.Launcher/App.cs
```diff
@@ -145,18 +142,13 @@ private void OnOutput(string s)
             }
         }
 
-        private bool HasVrMod(IList<IModData> mods)
```
amazingalek, line 148: I liked this, you could keep it :)

PR #147, opened 2020-07-04
Config menu for OWML

amazingalek:


OWML.ModLoader/ModLoader.cs
```diff
@@ -34,7 +35,8 @@ public static void LoadMods()
             var harmonyHelper = new HarmonyHelper(logger, console);
             var events = new ModEvents(logger, console, harmonyHelper);
             var inputHandler = new ModInputHandler(logger, console, harmonyHelper, owmlConfig, events);
-            var menus = new ModMenus(console, events, inputHandler);
+            var owmlData = new ModData(owmlManifest, null, null);
```
amazingalek, line 38: This is horrible, but can't think of a nice way to avoid it

OWML.ModHelper.Menus/ModsMenu.cs
```diff
@@ -54,6 +54,10 @@ public void Initialize(IModMainMenu mainMenu)
             var modsMenu = CreateModsMenu(optionsMenu);
             modsButton.OnClick += () => modsMenu.Open();
             Menu = mainMenu.Menu;
+
+            InitConfigMenu(_menus.OwmlMenu, optionsMenu);
+            var owmlButton = modsButton.Duplicate("OWML");
```
amazingalek, line 59: Maybe call it something else, like "Mod Loader" or "Mod Loader Settings"?

OWML.ModHelper.Menus/OwmlConfigMenu.cs
```diff
@@ -0,0 +1,53 @@
+﻿using OWML.Common;
+
+namespace OWML.ModHelper.Menus
+{
+    public class OwmlConfigMenu : ModConfigMenu
+    {
+        private const string GamePathTitle = "Game path";
+        private const string VerboseModeTitle = "Verbose mode";
+        private const string BlockInputTitle = "Block combination input";
```
amazingalek, line 9: Is this right, @TAImatem ?

OWML.ModLoader/ModLoader.cs
amazingalek: The mod data object is used by the base config menu which I override for the OWML config menu.
For the OWML config menu, mod data is only used once, in the constructor of the base class, which uses the manifest only.
The ways I've thought of refactoring it would lead to all sorts of side effects.

OWML.ModHelper.Menus/OwmlConfigMenu.cs
TAImatem: maybe rather "Block input on combination" or smth like that?

OWML.ModLoader/ModLoader.cs
TAImatem: well... It's not like you can't just add '=null' to contstructor itself or create a separate constructor

OWML.ModHelper.Menus/OwmlConfigMenu.cs
amazingalek: Does combination mean collision - "Block input on collision"?

OWML.ModLoader/ModLoader.cs
amazingalek: This is an attempt: https://github.com/amazingalek/owml/pull/148

OWML.Common/Menus/IModConfigMenu.cs
```diff
@@ -2,9 +2,12 @@
 {
     public interface IModConfigMenu : IModPopupMenu
     {
-        IModData ModData { get; }
```
amazingalek, line 5: This is a breaking change but I think no mods used it. Did any of you?

OWML.ModLoader/ModData.cs
```diff
@@ -17,15 +15,5 @@ public ModData(IModManifest manifest, IModConfig config, IModConfig defaultConfi
             DefaultConfig = defaultConfig;
         }
 
-        public void ResetConfig()
```
amazingalek, line 20: Removing this is a breaking change but I think no mods used it. Did any of you?

OWML.ModHelper.Menus/OwmlConfigMenu.cs
TAImatem: this option is for InputHandler's function for blocking game's InputCommands when combination is detected

Raicuparta: @TAImatem @amazingalek In Mod Manager I called it "Mod button combinations block game input". I think that makes it pretty clear, but maybe too long of a text for the game's menus?

amazingalek:
https://github.com/amazingalek/owml/pull/153 first

OWML.Common/Menus/IModConfigMenu.cs
amazingalek: Nevermind, re-added it

OWML.ModHelper.Menus/OwmlConfigMenu.cs
amazingalek: Surprisingly it fits :)

OWML.ModLoader/ModData.cs
amazingalek: I actually don't *have* to remove this any more, after introducing the base class... kinda wanna do it still 🤷

OWML.ModHelper.Menus/BaseConfigMenu.cs
```diff
@@ -0,0 +1,183 @@
...
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace OWML.ModHelper.Menus
+{
+    public abstract class BaseConfigMenu : ModPopupMenu, IBaseConfigMenu
+    {
+        public IModManifest Manifest { get; }
+
+        private readonly IModConsole _console;
```
TAImatem, line 15: please adopt changes I made in my PR (that is make console `protected` in ModMenu)

TAImatem: Kinda weird why you made it private in a first place. Almost every tier had it's own `_console`

OWML.ModHelper.Menus/ModComboInput.cs
```diff
@@ -39,10 +39,9 @@ public ModComboInput(TwoButtonToggleElement element, IModMenu menu, IModInputMen
 
         private void UpdateLayout(string currentCombination)
         {
-            var childCount = _layoutGroup.transform.childCount;
-            for (var i = childCount - 1; i >= 0; i--)
+            foreach (Transform child in _layoutGroup.transform)
             {
-                GameObject.Destroy(_layoutGroup.transform.GetChild(i).gameObject);
+                GameObject.Destroy(child.gameObject);
```
TAImatem, line 44: please don't. That's just you making things annoying either to me or to yourself. this part is moved to LayoutManager and tidied up there

OWML.ModHelper.Menus/ModsMenu.cs
TAImatem: I would say "Settings" is excessive. naming it "Mod Loader" sounds like a good idea though

OWML.ModLoader/ModData.cs
TAImatem: I suppose resetting to defaults is handled by `ModConfig` now?

amazingalek: Yeah, which do you think should do it - ModData or ModConfigMenu?

OWML.ModHelper.Menus/BaseConfigMenu.cs
amazingalek: Totally agree

amazingalek: I won't adopt those same changes in this PR tho, they'll be merged together to dev.

OWML.ModHelper.Menus/ModComboInput.cs
amazingalek: I don't agree that it's unnecessary but I get that it's annoying since it conflicts with your PR, reverting it 👍

OWML.ModHelper.Menus/ModsMenu.cs
amazingalek: Just tested with "Mod Loader". I tried thinking as a player, and when I saw it I expected it to be something else, maybe a way to load mods. So I reverted it to "OWML" again... 🤷‍♂️

OWML.ModLoader/ModData.cs
TAImatem: I think it depends whether or not you want to implement "Reset all mods to defaults" feature

TAImatem:
Other than what I mentioned things look ok to me

OWML.ModLoader/ModData.cs
amazingalek: Don't really see the point of that feature... do you?

TAImatem: not really...
BTW, when an item gets removed/added to default config do you automatically push that change to editable config?

TAImatem: Only reason I see is if something goes wrong after some updated mod gets confused by outdated config and user can't locate which one it is

amazingalek: If a value is added to the "settings" list/dict in default config, that isn't in config, it'll be added to config. Changes/removals to/of existing values won't be updated to config, until the user resets to default.

amazingalek: If you think there's value in keeping the reset method in mod data, I'll revert it. I don't have a strong opinion on this. Give the word :)

Raicuparta:
Now that we have a default OWML config, we shouldn't include `OWML.Config.json` in the releases, right? One of the main advantages of having a separate default config file is that new releases don't replace configs of previous releases (which can be very annoying if you need a custom game path).

createrelease.bat
```diff
@@ -7,6 +7,7 @@ copy "OWML.Patcher\bin\Debug\OWML.Patcher.dll" "Release\OWML.Patcher.dll"
 copy "OWML.ModLoader\bin\Debug\OWML.ModLoader.dll" "Release\OWML.ModLoader.dll"
 copy "OWML.Launcher\bin\Debug\OWML.Launcher.exe" "Release\OWML.Launcher.exe"
 copy "OWML.Launcher\bin\Debug\OWML.Config.json" "Release\OWML.Config.json"
+copy "OWML.Launcher\bin\Debug\OWML.DefaultConfig.json" "Release\OWML.DefaultConfig.json"
```
Raicuparta, line 10: Now that we have a default OWML config, we shouldn't include OWML.Config.json in the releases, right? One of the main advantages of having a separate default config file is that new releases don't replace configs of previous releases (which can be very annoying if you need a custom game path). OMLW.Config.json should only be generated from the default, after OWML starts.

amazingalek: Yeah, totally 👍 
If we do this: before starting OWML the first time, how will you show/edit the config in MM?

Raicuparta: Same as for mods. If mod manager sees only default, copy default to user config. Otherwise edit user config.

Need to make sure we copy each property and not only the whole file (in case we add new properties to config). Don't remember if mods already do this.

amazingalek: @Raicuparta removed config file, and creating it from default if it's missing.

amazingalek: Just copying the file, as there's no (reason to have a) list of generic settings.

OWML.Launcher/Program.cs
```diff
@@ -13,32 +13,44 @@ public class Program
     {
         static void Main(string[] args)
         {
-            var owmlConfig = GetOwmlConfig();
+            var owmlConfig = GetOwmlConfig() ?? CreateOwmlConfig();
```
Raicuparta, line 16: Won't this cause problems if you add new entries to the config? As in, if a user has a OWML.Config.json file, and then in a new update OWML adds a new configurable setting. The config exists, but is missing a property, so you're not getting that property from the default config.

amazingalek: I tried deleting a line from the config, to simulate your situation. It worked fine :-)
The config defaulted to the default of the data type (false for bool in this case) - it showed fine in the menu and saved properly after changing it.

Raicuparta: Hm but what if default was supposed to be true? 😅

OWML.ModLoader/ModData.cs
TAImatem: Only value I can see is making implementing "Reset All to Defaults" easier. And only reason I can see why user might want to reset all is if there are problems with outdated config and there is either multiple of them or user can't find which one is misbehaving

TAImatem: BTW, I suggest you to think about picking up type changes from default config (i.e. number to slider or to enum)

TAImatem: And warning on (seemingly) incompatible types

OWML.Launcher/Program.cs
amazingalek: Hmmm... 🤔 
I don't know how to code what you're suggesting without making a manually maintained mess. Any ideas?

OWML.ModLoader/ModData.cs
amazingalek: That sounds hard. Personally I don't think it would be worth it. But either way I'd argue it's outside the scope of this PR :)

OWML.Launcher/Program.cs
amazingalek: I guess iterating over properties using reflection. Sounds hard :P

Raicuparta: I guess too many things would have to change. What makes most sense for me is not to copy properties selectively, but to do in the moment of getting the property. So when you're trying to get OwmlConfig.GamePath, the getter checks if the property exists in OWML.Config.json, and if not, gets it from OWML.DefaultConfig.json.

Ideally, you wouldn't even need to create the OWML.Config.json file until the user changes a setting. And when they do, the user config would only need to include the settings that differ from the default.

But I understand that this can be to big of a change, and might not be easy to do with the way the JSON serialization is done (I think it always writes the whole file from the properties).

In any case, it's at the very least something to keep in mind if we add new settings ;)

PR #157, opened 2020-07-09
gathered all file reading/writing to one place

amazingalek:


OWML.ModHelper/ModStorage.cs
```diff
@@ -1,52 +1,26 @@
...
-                return default;
-            }
-            try
-            {
-                var json = File.ReadAllText(path);
-                return JsonConvert.DeserializeObject<T>(json);
-            }
-            catch (Exception ex)
-            {
-                _console.WriteLine($"Error while loading {path}: {ex}");
```
amazingalek, line 33: With this change, the error messages are lost

amazingalek: I guess we'll still get them with verbose mode enabled (which should be the new default soon)

Raicuparta: You can have a better error message in JSONHelper.cs though, no? Or would that be pointless with verbose mode?

amazingalek: Don't have a console, and this is (currently) static.
Could use ModConsole.Instance (obsolete/deprecated) but even that isn't always available (before manifest is loaded).

amazingalek: Not sure if verbose mode is enough, couldn't think of an easy way of testing it.

amazingalek:
Will change base branch to dev after https://github.com/amazingalek/owml/pull/147 is merged

PR #158, opened 2020-07-09
Input fields can be opened with controller

amazingalek:
I hate it but it works...

### DIFF
--- a/OWML.Common/Menus/IModButton.cs
+++ b/OWML.Common/Menus/IModButton.cs
@@ -23,5 +23,6 @@ namespace OWML.Common.Menus
         void Hide();
 
         void SetControllerCommand(SingleAxisCommand inputCommand);
+        void SetControllerCommand(SingleAxisCommand inputCommand, Selectable selectable);
     }
 }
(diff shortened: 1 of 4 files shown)


amazingalek:
This opens the text input popups with a controller, but doesn't let you close them (with a controller). Should that be implemented as well, @Raicuparta ?

Raicuparta:
> This opens the text input popups with a controller, but doesn't let you close them (with a controller). Should that be implemented as well, @Raicuparta ?

I think this is pretty important yes... It might seem like the game froze if suddenly nothing is responding to inputs controller inputs. Especially in situations where you don't immediately think to test the keyboard, like if you're playing on a TV or in VR.

TAImatem:
> 
> 
> This opens the text input popups with a controller, but doesn't let you close them (with a controller). Should that be implemented as well, @Raicuparta ?

Why doesn't it let you close it? Are you sure you setup popup correctly?

amazingalek:
> 
> 
> > This opens the text input popups with a controller, but doesn't let you close them (with a controller). Should that be implemented as well, @Raicuparta ?
> 
> Why doesn't it let you close it? Are you sure you setup popup correctly?

The popup doesn't support controller by default, haven't added support for it in OWML.

amazingalek:
Closing this, will use CommandListener instead of ControllerButton. New PR coming soon-ish.

PR #170, opened 2020-07-10
made Console protected in menus

TAImatem:
Not really sure what to do with empty .ctors


TAImatem:
TBF making Layout protected in this PR is kinda pointless since it's only used in ModInputCombinationMenu

PR #171, opened 2020-07-11
isSelected and common parents

TAImatem:
added IsSelected to ModInput and ModButton
added common parents where it made sense (Not sure if I should also implement class ModInputBase, common interface sounds enough for storage purposes)

### DIFF
--- a/OWML.Common/Menus/IModInput.cs
+++ b/OWML.Common/Menus/IModInput.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using UnityEngine;
 
 namespace OWML.Common.Menus
 {
--- a/OWML.Common/Menus/IModInput.cs
+++ b/OWML.Common/Menus/IModInput.cs
@@ -3,15 +3,9 @@ using UnityEngine;
 
 namespace OWML.Common.Menus
 {
-    public interface IModInput<T>
+    public interface IModInput<T> : IModInputBase
     {
         event Action<T> OnChange;
         T Value { get; set; }
-        MonoBehaviour Element { get; }
-        string Title { get; set; }
-        int Index { get; set; }
-        void Show();
-        void Hide();
-        void Initialize(IModMenu menu);
     }
 }
--- a/OWML.ModHelper.Menus/ModButton.cs
+++ b/OWML.ModHelper.Menus/ModButton.cs
@@ -23,7 +23,7 @@ namespace OWML.ModHelper.Menus
                 Button.transform.SetSiblingIndex(value);
             }
         }
-        public bool IsSelected => _uIStyleApplier.GetValue<bool>("_selected");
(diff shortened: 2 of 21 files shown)


TAImatem:
IsSelected is tested only for ModInputCombinationElement

TAImatem:
Although it kinda implies it would also work for all TwoButtonToggle-based things
Not yet sure if UIStyleApplier-based IsSelected works flawlessly

OWML.ModHelper.Menus/ModButton.cs
```diff
@@ -22,6 +23,17 @@ public int Index
                 Button.transform.SetSiblingIndex(value);
             }
         }
+        public bool IsSelected => _uIStyleApplier.GetValue<bool>("_selected");
```
amazingalek, line 26: can _uIStyleApplier be null? If so, this'll probably do:
`_uIStyleApplier?.GetValue<bool>("_selected") ?? false;`

OWML.Common/Menus/IModInput.cs
```diff
@@ -3,15 +3,9 @@
 
 namespace OWML.Common.Menus
 {
-    public interface IModInput<T>
+    public interface IModInput<T> : IModInputBase
```
amazingalek, line 6: This seems to be the only place that inherits from IModInputBase, so why should IModInput and IModInputBase be split?

amazingalek:
Super cool stuff! :)
Not sure we have the best naming of all of these UI elements tho ;)
Example: ModMenuInput, ModInputField - can we find more descriptive names?

OWML.ModHelper.Menus/ModButton.cs
TAImatem: ah, didn't think to use `??`

OWML.Common/Menus/IModInput.cs
TAImatem: because IModInput<bool> != IModInput<string>

OWML.ModHelper.Menus/ModButton.cs
TAImatem: Wait for sometime though I want to implement ModPromptButton in a separate branch first

TAImatem: done

OWML.Common/Menus/IModInput.cs
TAImatem: Idea is that you can do `Container<IModInputBase>`, but can't `Container<IModInput>`

TAImatem:
maybe ModFieldInput
Also ModPopupInput intead of ModMenuInput might be better

amazingalek:
Awesome.
You have a couple unused usings in interfaces (IModInput and IModInputBase)

PR #172, opened 2020-07-11
implemented modPromptButton

TAImatem:
doesn't really depend on IsSelected, but next PR will depend on both ;)


TAImatem:
Title is replaced with Prompt, kinda incompatible

TAImatem:
Inhereting from ModTitleButton might make it work shoddily 😝

OWML.Common/Menus/IModPromptButton.cs
```diff
@@ -0,0 +1,18 @@
...
+
+        new IModPromptButton Copy();
+        IModPromptButton Copy(string title);
+        IModPromptButton Copy(string title, int index);
+
+        IModPromptButton Duplicate(string title);
+        IModPromptButton Duplicate(string title, int index);
+
+        IModPromptButton Replace(string title);
+        IModPromptButton Replace(string title, int index);
```
amazingalek, line 16: I suggest leaving this out, until/unless you need them. In many cases I don't think we end up using these, and many UI elements don't have/need them. We have the base stuff anyways

OWML.ModHelper.Menus/ModPromptButton.cs
```diff
@@ -0,0 +1,89 @@
...
+            {
+                _prompt = value;
+                _hotkeyButton.SetPrompt(value);
+            }
+        }
+
+        public ModPromptButton(Button button, IModMenu menu) : base(button, menu)
+        {
+            _hotkeyButton = Button.GetComponent<ButtonWithHotkeyImageElement>();
+            _textID = Button.GetComponentInChildren<LocalizedText>(true).GetValue<UITextType>("_textID");
```
amazingalek, line 31: This doesn't have to happen if the null check below fails, so move it below the null check?

OWML.Common/Menus/IModPromptButton.cs
TAImatem: yeah, makes sense

OWML.ModHelper.Menus/ModPromptButton.cs
TAImatem: K

OWML.Common/Menus/IModPromptButton.cs
TAImatem: since it's basically only intended for sugarcoating existing buttons

TAImatem: and if someone wants to copy it (s)he would probably also want to set new Prompt

OWML.ModHelper.Menus/ModPromptButton.cs
```diff
@@ -0,0 +1,89 @@
...
+    {
+        private ScreenPrompt _prompt;
+        private UITextType _textID;
+        private readonly ButtonWithHotkeyImageElement _hotkeyButton;
+
+        public string DefaultTitle
+        {
+            get => UITextLibrary.GetString(_textID);
+        }
+        public ScreenPrompt Prompt
```
amazingalek, line 18: newline

TAImatem: Do I actually need newline between getters? pretty sure it's missing in other classes 😛

TAImatem: DefaultTitle can be turned into one-liner though

amazingalek: I've not seen any :O

amazingalek: Actually, single line getters can be close to each other, but not multiline ones ;)

amazingalek: Eh, it's fine, just resolve the conflicts and we're good ;)

OWML.Common/Menus/IModInputBase.cs
```diff
@@ -1,4 +1,4 @@
-﻿using UnityEngine;
+using UnityEngine;
```
amazingalek, line 1: Hmmm? :P

PR #186, opened 2020-07-12
sacrifice for a better code :P

TAImatem:
Ideally I'd also remove SetControllerButton, but that's technically part of API 🤷


OWML.ModHelper.Menus/ModPromptButton.cs
```diff
@@ -21,7 +24,7 @@ public ScreenPrompt Prompt
                 _hotkeyButton.SetPrompt(value);
                 if (_prompt.GetText() != DefaultTitle)
                 {
-                    Object.Destroy(Button.GetComponentInChildren<LocalizedText>());
+                    UnityEngine.Object.Destroy(Button.GetComponentInChildren<LocalizedText>());
```
TAImatem, line 27: apparently `using System` and `Object` don't like eachother

amazingalek: do `using Object = UnityEngine.Object;`, then you can do `Object.Destroy`

amazingalek:
> 
> 
> Ideally I'd also remove SetControllerButton, but that's technically part of API 🤷
It's probably only used by OWML, not by mods, so I'd say kill that too. @misternebula did you use it in your mods?

OWML.ModHelper.Menus/ModPromptButton.cs
```diff
@@ -1,7 +1,7 @@
 ﻿using OWML.Common.Menus;
 using OWML.ModHelper.Events;
 using UnityEngine.UI;
-using UnityEngine;
+using Object = UnityEngine.Object;
```
amazingalek, line 4: oops, not needed anymore, can be simplified to `using UnityEngine;`

amazingalek: Last thing ;)

PR #191, opened 2020-07-13
Tai/input handler improvements

TAImatem:
Not necessarily for this version
Implements https://github.com/amazingalek/owml/projects/1#card-41552452


TAImatem:
Just tested with TAIcheat. Seems to work without any problem

PR #200, opened 2020-07-13
Fix duplicate mods and fix dependancy errors

misternebula:


OWML.ModLoader/Owo.cs
```diff
@@ -56,17 +56,27 @@ public void LoadMods()
...
                 {
-                    _logger.Log("Mod type is null, skipping");
+                    if (missingDependencyFlag)
+                    {
+                        _logger.Log("Mod is missing its dependencies, skipping");
+                    }
+                    else
+                    {
+                        _logger.Log("Mod type is null, skipping");
+                    }
```
amazingalek, line 79:
```
var errorMessage= missingDependencyFlag ? "Mod is missing its dependencies, skipping" : "Mod type is null, skipping";
_logger.Log(errorMessage);
```

TAImatem: I did it a bit differently 😝

OWML.ModLoader/Owo.cs
```diff
@@ -51,28 +52,36 @@ public void LoadMods()
...
+                if (modData.Config.Enabled)
                 {
-                    if (!modNames.Contains(dependency) && mod.Config.Enabled)
+                    var missingDependencies = modData.Manifest.Dependencies
+                        .Where(dependency => !modNames.Contains(dependency));
+                    foreach (var dependency in missingDependencies)
                     {
-                        _console.WriteLine($"Error! {mod.Manifest.UniqueName} needs {dependency}, but it's disabled!");
+                        _console.WriteLine($"Error! {modData.Manifest.UniqueName} needs {dependency}, " +
+                                           "but it's disabled/missing!");
```
misternebula, line 69: hm. not too sure on this one. it reduces line length, but it puts a + in a place when you wouldn't expect it

OWML.ModLoader/Owo.cs
```diff
@@ -51,28 +52,36 @@ public void LoadMods()
...
-            foreach (var mod in sortedMods)
+            foreach (var modData in sortedMods)
             {
-                foreach (var dependency in mod.Manifest.Dependencies)
+                var isMissingDependency = false;
+                if (modData.Config.Enabled)
                 {
-                    if (!modNames.Contains(dependency) && mod.Config.Enabled)
+                    var missingDependencies = modData.Manifest.Dependencies
+                        .Where(dependency => !modNames.Contains(dependency));
```
misternebula, line 65: if dependencies doesn't contain anything (can't remember if it would be null in that case?) would this break?

OWML.ModLoader/Owo.cs
```diff
@@ -51,28 +52,36 @@ public void LoadMods()
...
                 }
-                var modType = LoadMod(mod);
-                if (modType == null)
+                var modType = LoadMod(modData);
+                if (modType == null || isMissingDependency)
                 {
-                    _logger.Log("Mod type is null, skipping");
-                    _menus.ModsMenu.AddMod(mod, null);
+                    _logger.Log(isMissingDependency ?
+                        "Mod is missing its dependencies, skipping" :
```
misternebula, line 77: this missing dependency message would be printed twice (this one and the one above)

OWML.ModLoader/Owo.cs
TAImatem: it would be empty I think. I didn't get any errors while testing with CustomLoader enabled

OWML.ModLoader/Owo.cs
amazingalek: Yep. How would you want it instead?

OWML.ModLoader/Owo.cs
misternebula: yeah, dependencies defaults to an empty string array, so this should be good 👍

OWML.ModLoader/Owo.cs
TAImatem: this only gets printed to log (like when mod is disabled it would print "Mod type is null"). TBH I'm not sure if that `.Log` call is needed at all

OWML.ModLoader/Owo.cs
TAImatem: not sure what's your problem with it

OWML.ModLoader/Owo.cs
amazingalek: It defaults to empty list: 
image

OWML.ModLoader/Owo.cs
amazingalek: You'd rather have it long?

misternebula: well normally +s are used to concat strings with variables, but this is meant to be one string. i'm pretty sure you can split a string into multiple lines without concat-ing, but it looks ugly 🤷‍♂️
i personally don't mind long lines, but ive proven myself to be the less experienced person here 😄

OWML.ModLoader/Owo.cs
amazingalek: Right. Stuff's printed in LoadMod and in the dependency check. Can probably be removed 👍 
I don't care enough to do it, tho ;)

misternebula: that first console would also get printed to the log though, and it actually names the mod in it... i don't think this log call is needed

OWML.ModLoader/Owo.cs
amazingalek: You're right, multiline strings exist in C#, and can be combined with string interpolation. Pushed.

OWML.ModLoader/Owo.cs
amazingalek: removed

PR #202, opened 2020-07-14
Tai/fix combination menu

TAImatem:
that seems to fix it

### DIFF
--- a/OWML.ModHelper.Menus/ModsMenu.cs
+++ b/OWML.ModHelper.Menus/ModsMenu.cs
@@ -47,11 +47,14 @@ namespace OWML.ModHelper.Menus
             var modsButton = owMenu.OptionsButton.Duplicate(ModsButtonTitle);
             var options = owMenu.OptionsMenu;
 
-            var toggleTemplate = options.InputTab.ToggleInputs[0].Copy().Toggle;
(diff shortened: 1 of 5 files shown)


TAImatem:
Ideally I would like to simply not re-initialize the combination menu, but it would require a bit more work

OWML.ModHelper.Menus/ModsMenu.cs
```diff
@@ -47,19 +47,19 @@ public void Initialize(IModOWMenu owMenu)
             var modsButton = owMenu.OptionsButton.Duplicate(ModsButtonTitle);
             var options = owMenu.OptionsMenu;
 
+            var toggleTemplate = options.InputTab.ToggleInputs[0].Copy().Toggle;
+            var comboElementTemplate = new ModInputCombinationElement(toggleTemplate,
+                _menus.InputCombinationMenu, _menus.InputCombinationElementMenu, _inputHandler);
+            var rebindMenuTemplate = options.RebindingMenu.Copy().Menu;
+            _menus.InputCombinationMenu.Initialize(rebindMenuTemplate, comboElementTemplate);
+
```
TAImatem, line 55: I did it just in case to preserve previous exec. order

PR #203, opened 2020-07-14
CombinationMenu gets initialized only once

TAImatem:


OWML.ModHelper.Menus/ModInputCombinationMenu.cs
```diff
@@ -51,11 +53,24 @@ protected override void SetupButtons()
 
         public void Initialize(Menu menu, IModInputCombinationElement combinationElementTemplate)
         {
-            _combinationElementTemplate = combinationElementTemplate;
-
-            base.Initialize(menu);
+            if (Menu == null)
```
TAImatem, line 56: technically this null check is not required, since it's checked in in ModsMenu anyway, feel free to perform direct commit if you want to remove it

PR #209, opened 2020-07-15
Keep user config consistent with default configs

### LINKED ISSUE #208: Support settings changing type between mod versions
If there's a config.json present, that one will be used to define the type of a setting. We need to compare config.json with default-config.json, and if the types of the same setting are different, we reset that setting to default.

Related: settings that are deleted from default-config.json should also be deleted from config.json.

Basically, the mod menu should always be build from default-config.json only if it exists. Leftover stuff in config.json shouldn't show up in the menus.

Raicuparta:
Closes #208 

Prevents problems when a mod update changes the type of a setting. To achieve consistency between user config and default config, we do this:

1. Add missing default properties to user config (was already happening before this change);
2. Remove options that are in user config, but not in default config;
3. Reset options to default if type in user config is different than type in default config.

For step 1, I've moved some logic from `ModFinder` to `ModConfig`'s `AddMissingDefaults()` method.

### DIFF
--- a/OWML.ModHelper/ModConfig.cs
+++ b/OWML.ModHelper/ModConfig.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using OWML.Common;
@@ -80,12 +81,10 @@ namespace OWML.ModHelper
                 if (!defaultConfig.Settings.ContainsKey(setting.Key))
                 {
                     settingsCopy.Remove(setting.Key);
-                    continue;
                 }
                 else if (!IsSettingSameType(setting.Value, defaultConfig.Settings[setting.Key]))
                 {
                     settingsCopy[setting.Key] = defaultConfig.Settings[setting.Key];
-                    continue;
                 }
             }
             Settings = settingsCopy;
@@ -93,29 +92,15 @@ namespace OWML.ModHelper
 
         private void AddMissingDefaults(IModConfig defaultConfig)
         {
-            foreach (var defaultSetting in defaultConfig.Settings)
-            {
-                if (!Settings.ContainsKey(defaultSetting.Key))
-                {
-                    Settings.Add(defaultSetting.Key, defaultSetting.Value);
-                }
-            }
+            var missingSettings = defaultConfig.Settings.Where(s => !Settings.ContainsKey(s.Key)).ToList();
+            missingSettings.ForEach(setting => Settings.Add(setting.Key, setting.Value));
         }
 
         private bool IsSettingSameType(object settingValue1, object settingValue2)
         {
-            if (settingValue1.GetType() != settingValue2.GetType())
-            {
-                return false;
-            }
-            if (settingValue1 is JObject obj1 && settingValue2 is JObject obj2)
-            {
-                if ((string)obj1["type"] != (string)obj2["type"])
-                {
-                    return false;
-                }
-            }
-            return true;
+            return settingValue1.GetType() == settingValue2.GetType() &&
+                   (!(settingValue1 is JObject obj1) || !(settingValue2 is JObject obj2) ||
+                    (string)obj1["type"] == (string)obj2["type"]);
         }
     }
 }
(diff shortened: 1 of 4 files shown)


OWML.ModHelper/ModConfig.cs
```diff
@@ -65,5 +65,57 @@ public T GetSetting<T>(string key)
...
+
+            var settingsCopy = new Dictionary<string, object>(Settings);
+            foreach (var setting in Settings)
+            {
+                if (!defaultConfig.Settings.ContainsKey(setting.Key))
+                {
+                    settingsCopy.Remove(setting.Key);
+                    continue;
+                }
+                else if (!IsSettingSameType(setting.Value, defaultConfig.Settings[setting.Key]))
```
TAImatem, line 85: maybe add separate detection if type is similar to not override the value? (i.e. number -> slider)

OWML.ModLoader/ModFinder.cs
```diff
@@ -53,17 +53,12 @@ private IModData GetModData(IModManifest manifest)
...
-                {
-                    if (!config.Settings.ContainsKey(setting.Key))
-                    {
-                        config.Settings.Add(setting.Key, setting.Value);
-                    }
-                }
+                config.MakeConfigConsistentWithDefaults(defaultConfig);
             }
+
+
```
TAImatem, line 59: double newline? why?

OWML.ModHelper/ModConfig.cs
TAImatem: or string -> combination

OWML.ModHelper/ModConfig.cs
```diff
@@ -65,5 +65,57 @@ public T GetSetting<T>(string key)
...
+            Settings = settingsCopy;
+        }
+
+        private void AddMissingDefaults(IModConfig defaultConfig)
+        {
+            foreach (var defaultSetting in defaultConfig.Settings)
+            {
+                if (!Settings.ContainsKey(defaultSetting.Key))
+                {
+                    Settings.Add(defaultSetting.Key, defaultSetting.Value);
```
misternebula, line 100: pretty sure you can do something like
`Settings.Concat(defaultConfig.Settings.Where( x=> !defaultConfig.Settings.Keys.Contains(defaultSetting.Key)));`

TAImatem: maybe it's Union though

OWML.ModHelper/ModConfig.cs
Raicuparta: Sounds like something for another PR. Don't see the need to complicate this further.

OWML.ModHelper/ModConfig.cs
misternebula: i think unions don't support dicts without a custom comparer

OWML.ModHelper/ModConfig.cs
TAImatem: Sounds like it might require some manual coding though

OWML.ModHelper/ModConfig.cs
TAImatem: pretty sure I used Union in HashSets, don't think Dictionary would be much different

Raicuparta: That doesn't look any better to me. Just makes it hard to read with no obvious advantage.

I know @amazingalek has a sweet spot for this kind of one-liner, so maybe he can chime in.

OWML.ModLoader/ModFinder.cs
Raicuparta: good catch, would have thought that VS would auto-fix since I have it configured for that

OWML.ModHelper/ModConfig.cs
Raicuparta: Have you not been coding manual until now? 🎷

TAImatem: I mean it doesn't sound like it could be solved with single `if`

OWML.ModHelper/ModConfig.cs
Raicuparta: BTW this code was just moved over from `ModFinder.cs` with almost no changes.

TAImatem: ah, it's UnionWith rather

amazingalek: https://github.com/amazingalek/owml/pull/210 🥴

PR #211, opened 2020-07-15
constants for filenames

amazingalek:


amazingalek:
> good idea, even though my dislike of pascal case constants is telling me otherwise 😋

Feel free to argue against the convention, I'm sure you'll be very persuasive ;)

PR #212, opened 2020-07-15
updated config section in readme and changed wording of input config

amazingalek:


OWML.ModHelper.Menus/OwmlConfigMenu.cs
```diff
@@ -4,9 +4,8 @@ namespace OWML.ModHelper.Menus
 {
     public class OwmlConfigMenu : ModConfigMenuBase
     {
-        private const string GamePathTitle = "Game path";
         private const string VerboseModeTitle = "Verbose mode";
-        private const string BlockInputTitle = "Mod button combinations block game input";
+        private const string BlockInputTitle = "Mod input combinations block game input";
```
misternebula, line 8: This is still a little confusing... I haven't been keeping up with the combinations stuff, so I have no idea what this means

amazingalek: Totally. I've had it explained several times and I still don't quite get it. I tried to make it more understandable here, but not sure if I succeeded.

TAImatem: When it's on InputHandler would for example block `Shift` from game while registered combination `Shift+L` is being pressed

Raicuparta: Maybe we stop calling it "combinations"? It's trying to refer to the shortcuts that mods can define as a mod setting. As in, "if this setting is set to true, inputs used by mods will block the actions that the same shortcut would cause in the base game". For instance, if you had a shortcut that includes the jump button, pressing that shortcut would cause the player to jump, unless this option is enabled.

Some examples that might be clearer:
1. "Prevent game actions from being triggered by mod inputs"
2. "Mod inputs can block triggering of game actions"

amazingalek: 👍 I tried alt. 2

PR #218, opened 2020-07-18
Implemented selector input

TAImatem:


OWML.ModHelper.Menus/ModConfigMenuBase.cs
```diff
@@ -81,6 +84,11 @@ protected void AddConfigInput(string key, object value, int index)
                     AddToggleInput(key, obj, index);
                     return;
                 }
+                if (type == "selector")
+                {
+                    AddSelectorInput(key, obj, index);
+                    return;
+                }
```
misternebula, line 91: Could this be a switch case?

TAImatem: that's out of scope  of this PR tho

OWML.SampleMods/OWML.LoadCustomAssets/LoadCustomAssets.cs
```diff
@@ -48,6 +54,9 @@ public override void Configure(IModConfig config)
             var speed = config.GetSettingsValue<float>("speed");
             var power = config.GetSettingsValue<float>("power");
             var enableSuperMode = config.GetSettingsValue<bool>("enableSuperMode");
+            var selectedIndex = config.GetSettingsValue<ABC>("thing");
```
Raicuparta, line 57: forgot to rename to `selectedEnum`?

Also, include an example of a setting with numeric options.

OWML.ModHelper/ModConfig.cs
```diff
@@ -30,12 +30,20 @@ public T GetSettingsValue<T>(string key)
...
+                    return Enum.IsDefined(type, thisEnum) ? (T)thisEnum : default;
+                }
+                var val = objectValue != null ? objectValue["value"] : value;
+                return (T)Convert.ChangeType(val, type);
             }
-            catch (InvalidCastException)
+            catch (InvalidCastException ex)
             {
-                ModConsole.Instance.WriteLine($"Error when converting setting {key} of type {value.GetType()} to type {typeof(T)}");
+                ModConsole.Instance.WriteLine($"Error when converting setting {key} of type {value.GetType()} to type {typeof(T)} : {ex.Message};{ex.TargetSite}");
```
Raicuparta, line 46:
```suggestion
                ModConsole.Instance.WriteLine($"Error when converting setting {key} of type {value.GetType()} to type {typeof(T)} : {ex}");
```
Converting the exception to string already includes Message and TargetSite (and more stuff).

OWML.Common/Menus/IModConfigMenuBase.cs
```diff
@@ -5,6 +5,7 @@ public interface IModConfigMenuBase : IModPopupMenu
         IModManifest Manifest { get; }
 
         void Initialize(Menu modMenuCopy, IModToggleInput toggleTemplate, IModSliderInput sliderTemplate, 
-            IModTextInput textInputTemplate, IModNumberInput numberInputTemplate, IModComboInput comboInputTemplate);
+            IModTextInput textInputTemplate, IModNumberInput numberInputTemplate,
+            IModComboInput comboInputTemplate, IModSelectorInput selectorTemplate);
```
Raicuparta, line 9: If we get to a point where we need to have line breaks between method arguments, we should always have only one argument per line, or it becomes awkward to read (especially in diffs).

I know this example was already split before, but it's a good chance to clean it up.

I think @amazingalek agreed with me before on this, but maybe that was on the QSB codebase? What do you think.

OWML.ModHelper/ModConfig.cs
TAImatem: ah, I simply forgot to remove that thing >.<

Raicuparta: It's much better to include the exception though. There's a few places in OWML that we consume the exception and it makes for hard debugging.

TAImatem: in this case it won't tell much AFAIK

TAImatem: AFAIK it would simply say "can't convert from source to target" which is even less informative than OWML's error message

Raicuparta: I see. But it would still have info related to which method threw the error, which is useful.

TAImatem: I wasn't able to make it show that :/

Raicuparta: Oh, I thought Exception.ToString would always do that. Do whatever you please then 🤷‍♂️

OWML.SampleMods/OWML.LoadCustomAssets/LoadCustomAssets.cs
```diff
@@ -48,6 +54,10 @@ public override void Configure(IModConfig config)
             var speed = config.GetSettingsValue<float>("speed");
             var power = config.GetSettingsValue<float>("power");
             var enableSuperMode = config.GetSettingsValue<bool>("enableSuperMode");
+            var selectedEnum = config.GetSettingsValue<ABC>("thing");
+            var selectedString = config.GetSettingsValue<string>("thing");
```
amazingalek, line 58: seems a bit excessive to support both string and enum at the same time?

OWML.ModHelper/ModConfig.cs
```diff
@@ -30,8 +30,16 @@ public T GetSettingsValue<T>(string key)
...
+                var objectValue = value as JObject;
+                var type = typeof(T);
+                var val = (objectValue != null) ? objectValue["value"] : value;
+                if (type.IsEnum && objectValue != null && objectValue["type"].ToObject<string>() == "selector")
+                {
+                    var selected = Convert.ToString(val);
+                    var thisEnum = Enum.Parse(type, selected, true);
+                    return Enum.IsDefined(type, thisEnum) ? (T)thisEnum : default;
+                }
+                return (T)Convert.ChangeType(val, type);
```
amazingalek, line 42: can this be simplified? maybe extract the selector stuff to new method? 
why is only the selector a special case here and nothing else?

OWML.SampleMods/OWML.LoadCustomAssets/LoadCustomAssets.cs
TAImatem: String is it's native format, no reason to not support it 🤷‍♂️

OWML.ModHelper/ModConfig.cs
TAImatem: Mostly because only selector offers nice enum-like structure. Technically it would also work with any other string-based inputs, but it doesn't make much sense because of high likelihood of incorrect values

OWML.SampleMods/OWML.LoadCustomAssets/LoadCustomAssets.cs
Raicuparta: If we were to remove one we'd remove Enum, but I think both can be useful (if you're feeling lazy use string / int, if you want cleaner code with less magic strings use Enum).

OWML.Common/Menus/IModConfigMenuBase.cs
TAImatem: IMO it's kinda out of scope of this PR, possibly an extra code cleanup round by Alek/Me/You will do that 🤷‍♂️

OWML.ModHelper/ModConfig.cs
```diff
@@ -30,8 +30,16 @@ public T GetSettingsValue<T>(string key)
 
             try
             {
-                var val = value is JObject obj ? obj["value"] : value;
-                return (T)Convert.ChangeType(val, typeof(T));
+                var objectValue = value as JObject;
+                var type = typeof(T);
+                var val = (objectValue != null) ? objectValue["value"] : value;
```
amazingalek, line 35: seems this line is only used below the if?

amazingalek: ah, no, not quite. still feel like this could be simplified. hard for me to say for sure via phone 😋

OWML.SampleMods/OWML.LoadCustomAssets/LoadCustomAssets.cs
amazingalek: right, the enum is the special case. covered by other comments 👍

OWML.Common/Menus/IModConfigMenuBase.cs
Raicuparta: Boy scout rule again :) Small clean up (not big refactors) is ok when you're touching messy code.

OWML.ModHelper/ModConfig.cs
amazingalek: in other words: it's a bit confusing to read. 
maybe better variable names could help. 
maybe inverting the if would help.
maybe moving the "val" variable below the if and fixing the resulting error in the if would help.

amazingalek: and maybe extracting the enum thing to another method would help

OWML.ModHelper/ModConfig.cs
Raicuparta: I guess I'm partly at fault because I previously said it makes more sense to support only selector for enums. But I didn't realise it would actually simplify the code. The only difference would be the error the developer sees in case of failure. In this case we can just make the condition just `type.IsEnum` and remove the check for selector. 

Would still be good to have a separate method for converting to Enum though.

OWML.ModHelper/ModConfig.cs
Raicuparta: Would also be good to have less val / value variables, but I think that was already a problem in this settings code in general.

OWML.ModHelper/ModConfig.cs
TAImatem: Problem is it's impossible to do where T : enum :(

OWML.ModHelper/ModConfig.cs
```diff
@@ -26,16 +42,28 @@ public T GetSettingsValue<T>(string key)
...
+
+                if (type.IsEnum && value is float || value is double)
+                {
+                    var floatValue = Convert.ToDouble(value);
+                    value = (long)Math.Round(floatValue);
+                }
+
+                if (!type.IsEnum || value is int || value is long)
+                {
+                    return type.IsEnum? (T)value : (T)Convert.ChangeType(value, type);
```
amazingalek, line 60: you already checked type.IsEnum

OWML.ModHelper/ModConfig.cs
```diff
@@ -26,16 +42,28 @@ public T GetSettingsValue<T>(string key)
...
-                return (T)Convert.ChangeType(val, typeof(T));
+                var value = setting is JObject objectValue ? objectValue["value"] : setting;
+
+                if (type.IsEnum && value is float || value is double)
+                {
+                    var floatValue = Convert.ToDouble(value);
+                    value = (long)Math.Round(floatValue);
+                }
+
+                if (!type.IsEnum || value is int || value is long)
```
amazingalek: with the method extraction it would be better to un-invert the if again ;-P

OWML.ModHelper/ModConfig.cs
```diff
@@ -18,6 +19,31 @@ public class ModConfig : IModConfig
         [JsonProperty("settings")]
         public Dictionary<string, object> Settings { get; set; } = new Dictionary<string, object>();
 
+        private T ConvertToEnum<T>(object value)
```
amazingalek, line 22: private under public 🙂

OWML.ModHelper/ModConfig.cs
```diff
@@ -18,6 +19,31 @@ public class ModConfig : IModConfig
...
+        {
+            if (value is float || value is double)
+            {
+                var floatValue = Convert.ToDouble(value);
+                value = (long)Math.Round(floatValue);
```
amazingalek, line 27: why not return here, but return if int?

amazingalek: ah, i see. would prefer return here, at the cost of a few more (duplicated) words

amazingalek: also, should 4.6 become 5? i guess this question doesn't even make sense 😁

amazingalek: we're in *undefined* territory. i guess just casting (same as floor) is more intuitive tho?

OWML.ModHelper/ModConfig.cs
TAImatem: no longer valid :P

OWML.ModHelper/ModConfig.cs
TAImatem: no longer valid :P

OWML.ModHelper/ModConfig.cs
TAImatem: actually, yeah, I can just return here

OWML.ModHelper/ModConfig.cs
TAImatem: OK

OWML.ModHelper/ModConfig.cs
```diff
@@ -4,6 +4,7 @@
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using OWML.Common;
+using UnityEngine.EventSystems;
```
amazingalek, line 7: are you using this, and if yes: for what?

TAImatem: probably got added automatically, will remove that

OWML.ModHelper/ModConfig.cs
TAImatem: done that way because of slider

PR #219, opened 2020-07-19
replaced ifs in AddConfigInput with switch

TAImatem:


TAImatem:
gotta merge base first tho

amazingalek:
You can merge it now. Your other PR isn't too large.  This is just the boy scout rule: leave the place a little nicer than when you arrived. It's only an issue when the PR is already enormous ;-)

PR #220, opened 2020-07-20
Add message types and JSON

misternebula:


misternebula:
This should only be merged once socket-logging-only is merged, so i'll rebase when that's merged

OWML.Common/IModConsole.cs
```diff
@@ -1,8 +1,17 @@
...
+namespace OWML.Common
 {
     public interface IModConsole
     {
+        [Obsolete]
         void WriteLine(string s);
+        [Obsolete]
         void WriteLine(params object[] s);
+
+        void WriteLine(string sender, MessageType type, string s);
```
amazingalek, line 12: Mods and OWML have their own console/logging objects initialized with manifests, to avoid giving sender explicitly when logging.
It doesn't feel right that when i log in my mod I can lie about who I am this easily.

amazingalek: In other words: remove sender parameter, take it from manifest instead

OWML.Common/MessageType.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace OWML.Common
+{
+    public enum MessageType
```
amazingalek, line 8: LogType?

OWML.Common/MessageType.cs
```diff
@@ -0,0 +1,15 @@
...
+    {
+        Error,
+        Warning,
+        Message,
```
amazingalek, line 12: Info?

OWML.Common/MessageType.cs
```diff
@@ -0,0 +1,15 @@
...
+        Success
```
amazingalek, line 13: Not usually a log level but I kinda like it

OWML.Launcher/SocketListener.cs
```diff
@@ -70,15 +69,32 @@ private void ListenToSocket(ref TcpListener server)
...
         private void PrintOutput(byte[] bytes, int count)
         {
-            var data = Encoding.ASCII.GetString(bytes, 0, count);
-            var objects = data.Split(new string[] { ";;" }, StringSplitOptions.None);
+            var json = Encoding.ASCII.GetString(bytes, 0, count);
+
+            var data = JsonConvert.DeserializeObject<SocketMessage>(json);
 
-            if (objects[1] == Constants.QuitKeyPhrase)
+            if (data.Message == Constants.QuitKeyPhrase)
```
amazingalek, line 76: I guess this magic phrase could be a message type instead...

OWML.Common/MessageType.cs
amazingalek: Actually, maybe message type makes sense - considering my comment about the magic quit phrase...

OWML.Common/MessageType.cs
Raicuparta: @amazingalek when would we use Info vs Message?

Raicuparta: Actually nevermind, info would be cool to have too.

OWML.Common/IModConsole.cs
misternebula: done

OWML.Common/MessageType.cs
misternebula: Added info with blue color

OWML.Launcher/SocketListener.cs
misternebula: done!

OWML.Common/MessageType.cs
```diff
@@ -10,6 +10,8 @@ public enum MessageType
         Error,
         Warning,
         Message,
-        Success
+        Info,
+        Success,
+        QuitPhrase
```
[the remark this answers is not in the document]
misternebula: done

OWML.Launcher/SocketListener.cs
```diff
@@ -27,9 +27,7 @@ private void SetupSocketListener()
             }
             catch (SocketException ex)
             {
-                Console.ForegroundColor = ConsoleColor.Red;
-                Console.WriteLine("Error in socket listener : " + ex);
-                Console.ForegroundColor = ConsoleColor.Gray;
+                ConsoleUtils.WriteLineWithColor(ConsoleColor.Red, $"Error in socket listener: {ex}");
```
Raicuparta, line 30: Maybe you should have a method in `SocketListener` just to log these errors, like
```
private void WriteToConsole(MessageType type, string message) {
    ConsoleUtils.WriteLineWithColor(ConsoleUtils.ConsoleColorFromMessageType(type), message);
}
```

Raicuparta: Or maybe you can drop WriteLineWithColor and just have a method that writes by type? Why would you need to manually specify a color?

OWML.ModHelper/Logging/ModSocketOutput.cs
```diff
@@ -46,26 +43,49 @@ public override void WriteLine(string s)
...
+            WriteLine(senderName, senderFile, type, s);
         }
 
         [Obsolete("Use ModSocketOutput.Writeline(MessageType type, params object[] objects) instead")]
         public override void WriteLine(params object[] objects)
         {
             if (CheckForParamsError(objects))
             {
-                WriteLine(string.Join(" ", objects.Select(o => o.ToString()).ToArray()));
+                var s = string.Join(" ", objects.Select(o => o.ToString()).ToArray());
```
Raicuparta, line 56: `s` needs a more descriptive name.

OWML.ModHelper/Logging/ModSocketOutput.cs
```diff
@@ -46,26 +43,49 @@ public override void WriteLine(string s)
             {
                 type = MessageType.Message;
             }
-            WriteLine(type, s);
+            var senderName = Manifest.Name;
+            var senderFile = (new System.Diagnostics.StackTrace()).GetFrame(1).GetMethod().DeclaringType.Name;
```
Raicuparta, line 47: You're doing this in two places. Write a method for it, and make sure it still works in case this throws an error (wrap in try catch and, in case of failure, return null or empty string or whatever works here).

OWML.Launcher/SocketListener.cs
misternebula: done

OWML.ModHelper/Logging/ModSocketOutput.cs
misternebula: Can't do it in a seperate method. that stacktrace needs to be in the exact method that was called

OWML.Common/SocketMessage.cs
```diff
@@ -7,7 +7,8 @@ namespace OWML.Common
 {
     public class SocketMessage
     {
-        public string Sender;
+        public string SenderName;
+        public string SenderFile;
```
Raicuparta, line 11: You get this string from `Method.DeclaringType.Name`, so maybe it should be called `SenderType` instead?

OWML.ModHelper/Logging/ModSocketOutput.cs
misternebula: that's the reason WriteLine(string) and WriteLine(params object[]) are now more complicated

misternebula: and why there's now a private WriteLine(string senderName, string SenderFile, MessageType type, string line)

OWML.Common/SocketMessage.cs
misternebula: done

OWML.ModHelper/Logging/ModSocketOutput.cs
misternebula: done

OWML.ModHelper/Logging/ModSocketOutput.cs
misternebula: fixed in latest commit

OWML.Common/ConsoleUtils.cs
```diff
@@ -0,0 +1,37 @@
...
+                    Console.ForegroundColor = ConsoleColor.Green;
+                    break;
+                case MessageType.Message:
+                    Console.ForegroundColor = ConsoleColor.Gray;
+                    break;
+                case MessageType.Info:
+                    Console.ForegroundColor = ConsoleColor.Blue;
+                    break;
+                case MessageType.Quit:
+                    Environment.Exit(0);
```
amazingalek, line 30: this is not what the method name suggests 😁

OWML.Common/IModConsole.cs
```diff
@@ -1,8 +1,12 @@
-﻿namespace OWML.Common
+﻿using System;
+
+namespace OWML.Common
 {
     public interface IModConsole
     {
-        void WriteLine(string s);
```
amazingalek, line 5: this breaks backwards compatibility?

OWML.Launcher/SocketListener.cs
```diff
@@ -70,16 +65,12 @@ private void ListenToSocket(ref TcpListener server)
 
         private void PrintOutput(byte[] bytes, int count)
         {
-            var data = Encoding.ASCII.GetString(bytes, 0, count);
-            var objects = data.Split(new string[] { ";;" }, StringSplitOptions.None);
+            var json = Encoding.ASCII.GetString(bytes, 0, count);
```
amazingalek, line 68: probably UTF8?

OWML.ModHelper/Logging/ModSocketOutput.cs
```diff
@@ -20,34 +20,91 @@ public ModSocketOutput(IOwmlConfig config, IModLogger logger, IModManifest manif
...
+            }
+            else if (line.ToLower().Contains("success"))
+            {
+                type = MessageType.Success;
+            }
+            else
+            {
+                type = MessageType.Message;
+            }
+            var senderFile = GetCallingMethodName(new StackTrace());
```
amazingalek, line 48: senderFile or senderName?

OWML.ModHelper/Logging/ModSocketOutput.cs
```diff
@@ -20,34 +20,91 @@ public ModSocketOutput(IOwmlConfig config, IModLogger logger, IModManifest manif
...
+            WriteLine(senderFile, type, line);
         }
 
-        public override void WriteLine(params object[] objects)
+
+        public override void WriteLine(string line, MessageType type = MessageType.Message)
```
amazingalek, line 53: Don't remember this in the interface

amazingalek: Can this method call the one below or opposite? Seems like they have duplicate code

OWML.ModHelper/OwmlBehaviour.cs
```diff
@@ -13,7 +13,7 @@ private void Start()
 
         private void OnApplicationQuit()
         {
-            ModConsole.Instance.WriteLine(Constants.QuitKeyPhrase);
+            ModConsole.Instance.WriteLine(MessageType.Quit, "RageAgainstTheDyingOfTheLight");
```
amazingalek, line 16: message is irrelevant here, right? so null or ""?

OWML.Common/ConsoleUtils.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace OWML.Common
+{
+    public static class ConsoleUtils
```
amazingalek, line 8: whats the reason for this class?
can't its method be part of ModConsole or one of its children?
if it must exist - must it be static?

OWML.ModHelper/Logging/ModSocketOutput.cs
```diff
@@ -20,34 +20,91 @@ public ModSocketOutput(IOwmlConfig config, IModLogger logger, IModManifest manif
...
+                SenderType = senderFile,
+                Type = type,
+                Message = line
+            };
+            var json = JsonConvert.SerializeObject(message);
+
+            WriteToSocket(json);
+        }
+
+        private string GetCallingMethodName(StackTrace frame)
```
amazingalek, line 87: very cool 👍

OWML.ModHelper/Logging/ModSocketOutput.cs
Raicuparta: Actually should be `senderType`. There's a few other places with the wrong name still.

OWML.Common/ConsoleUtils.cs
misternebula: this code was duplicated in program, socketlistener and outputwriter

misternebula: isn't it easier to have a static class you can just use than making a `new` every time you want to use it?

OWML.Common/IModConsole.cs
misternebula: shouldnt do, as the new method is `void WriteLine(string line, MessageType type = MessageType.Message);`
(Admittedly I didn't have the `= MessageType.Message` in there when you reviewed, so that's on me :P)

OWML.ModHelper/Logging/ModSocketOutput.cs
misternebula: done 👍

OWML.Common/IModConsole.cs
amazingalek: i've tried adding default parameters to existong methods before, it breaks backwards compatibility, even if it's not obvious

OWML.Common/ConsoleUtils.cs
amazingalek: it's great that you collected duplicated code to one place 👍
looks like all places this is used is from Launcher? in that case, consider putting this file there.
also, looks like this code can be put inside OutputWriter.cs? i know it's not in Launcher, but i guess it's only used from there - if so you can consider doing both: put this code inside OutputWriter and move it to Launcher. 
hard to properly check all this from phone 😋

amazingalek: hypothesis: this belongs in OutputWriter.WriteLine. it must be newed up early and used (re-used - the same object) all the places you identified. i'm unable to verify that this makes 100% sense

OWML.Common/IModConsole.cs
misternebula: well all the OWML log messages, and the ones from the example mods work

OWML.Common/ConsoleUtils.cs
misternebula: it's only using in program and socketlistener for writing errors/connection status to the console... could be removed?

OWML.Common/IModConsole.cs
amazingalek: yeah but with old mods?

OWML.ModHelper/Logging/ModSocketOutput.cs
```diff
@@ -20,34 +20,79 @@ public ModSocketOutput(IOwmlConfig config, IModLogger logger, IModManifest manif
...
+        [Obsolete]
+        public override void WriteLine(params object[] objects)
         {
-            Logger?.Log(s);
-            CallWriteCallback(Manifest, s);
-            var message = $"{Manifest.Name}{NameMessageSeparator}{s}";
-            InternalWriteLine(message);
+            var line = string.Join(" ", objects.Select(o => o.ToString()).ToArray());
+            MessageType type;
+            if (line.ToLower().Contains("error") || line.ToLower().Contains("exception"))
```
amazingalek, line 32: not you who created this horrible if-else, but: 
you could extract one of the ifs to a new private method and simplify it:
```
private bool Contains(string line, params string[] keyWords)
{
    return keyWords.Any(keyWord => line.Contains);
}
```

amazingalek: After doing that, maybe VS will suggest something even shorter

OWML.ModHelper/Logging/ModSocketOutput.cs
```diff
@@ -20,34 +20,79 @@ public ModSocketOutput(IOwmlConfig config, IModLogger logger, IModManifest manif
...
 
-        public override void WriteLine(params object[] objects)
+
+        public override void WriteLine(string line, MessageType type = MessageType.Message)
         {
-            WriteLine(string.Join(" ", objects.Select(o => o.ToString()).ToArray()));
+            WriteLine(GetCallingMethodName(new StackTrace()), type, line);
+        }
+
+        private void WriteLine(string senderType, MessageType type, string line)
```
amazingalek, line 58: the parameter order is inconsistent between the different methods.
I suggest: line, type, sender.

OWML.Common/IModConsole.cs
TAImatem: that's because they got recompiled

misternebula: fixed 👍

OWML.ModHelper/Logging/ModSocketOutput.cs
misternebula: fixed i think?

OWML.ModHelper/Logging/ModSocketOutput.cs
amazingalek: and do `var type = GetType(line);`

OWML.ModHelper/Logging/ConsoleUtils.cs
```diff
@@ -0,0 +1,65 @@
...
+                    Console.ForegroundColor = ConsoleColor.Cyan;
+                    break;
+            }
+            Console.WriteLine(line);
+            Console.ForegroundColor = ConsoleColor.Gray;
+        }
+
+        public static MessageType ContentsToType(string line)
+        {
+            if (ErrorList.Contains(line.ToLower()))
```
amazingalek, line 50: if you do this:

```
private bool Contains(string line, params string[] keyWords)
{
return keywords.Any(keyWord => line.Contains);
}
```
and use it like this:
`if (Contains(line, "error", "warning"))`
then you can get rid of the lists

OWML.ModHelper/Logging/ConsoleUtils.cs
```diff
@@ -0,0 +1,65 @@
...
+                    break;
+                case MessageType.Info:
+                    Console.ForegroundColor = ConsoleColor.Cyan;
+                    break;
+            }
+            Console.WriteLine(line);
+            Console.ForegroundColor = ConsoleColor.Gray;
+        }
+
+        public static MessageType ContentsToType(string line)
```
amazingalek, line 48: if this method is only used one place there's no need to move it from that place?

amazingalek: Actually, this is another reason i hate static classes. They're accessable from anywhere, so they tend to be used from anywhere, and gradually become trash cans for random code that don't really belong together. It's a slippery slope, you won't believe the things I've seen 😋
So - if this method is only used from one place - move it (back) to that place ;-)

PR #222, opened 2020-07-21
Improvements upon Raicuparta's config fixer

TAImatem:


TAImatem:
Found an issue, would need some testin' and fixin'

OWML.ModHelper/ModConfig.cs
```diff
@@ -104,27 +108,82 @@ public void MakeConsistentWithDefaults(IModConfig defaultConfig)
...
-                    settingsCopy[setting.Key] = defaultConfig.Settings[setting.Key];
+                    UpdateSelector(key, Settings[key], objectValue);
                 }
             }
-            Settings = settingsCopy;
+
+            AddMissingDefaults(defaultConfig);
+        }
+
+        private bool UpdateSelector(string key, object userSetting, JObject modderSetting)
```
amazingalek, line 130: return value isn't used, and removing it would make it cleaner

OWML.ModHelper/ModConfig.cs
```diff
@@ -104,27 +108,82 @@ public void MakeConsistentWithDefaults(IModConfig defaultConfig)
...
+            return false;
         }
 
         private void AddMissingDefaults(IModConfig defaultConfig)
         {
             var missingSettings = defaultConfig.Settings.Where(s => !Settings.ContainsKey(s.Key)).ToList();
             missingSettings.ForEach(setting => Settings.Add(setting.Key, setting.Value));
+        }   
+
+        private bool TryUpdate(string key, object userSetting, object modderSetting)
```
amazingalek, line 149: this method is very long.
if you return in the ifs instead of changing state, it would be easier to break it up into smaller methods

OWML.ModHelper/ModConfig.cs
TAImatem: that's a setup for warning PR

OWML.ModHelper/ModConfig.cs
TAImatem: I don't think it's possible to return in ifs

TAImatem: I feel like moving `Settings[key] = modderSetting;` to start might affect userSettings, imma test that

TAImatem: alright, doesn't seem to be the case. Shortened TryUpdate a little bit

OWML.ModHelper/ModConfig.cs
```diff
@@ -104,27 +108,79 @@ public void MakeConsistentWithDefaults(IModConfig defaultConfig)
...
+        private bool UpdateSelector(string key, object userSetting, JObject modderSetting)
+        {
+            var options = modderSetting["options"].ToObject<List<string>>();
+            var userString = userSetting is JObject objectValue ? (string)objectValue["value"] : Convert.ToString(userSetting);
+            Settings[key] = modderSetting;
+            if (options.Contains(userString))
+            {
+                SetSettingsValue(key, userString);
+                return true;
+            }
```
amazingalek, line 139:
```
var containsUserString = options.Contains(userString);
if (containsUserString) {
SetSettingsValue(key, userString);
}
return containsUserString;
```
saved one line 😋

TAImatem: OK

PR #224, opened 2020-07-22
Show slider value in mod settings

### LINKED ISSUE #223: Show slider value in mod settings

Raicuparta:
There does't seem to be a way to just toggle the display of the value inside the slider handle; seems like the elements are created in the editor, and then the text has to be updated manually (which is kinda crazy).

Had to change the slider template since the one being used before didn't include this text inside the handle.

I made it so every slider always shows the currently selected value. Could make it optional, but I felt like it was unnecessary.

Closes #223

### DIFF
--- a/OWML.ModHelper.Menus/ModSliderInput.cs
+++ b/OWML.ModHelper.Menus/ModSliderInput.cs
@@ -54,9 +54,14 @@ namespace OWML.ModHelper.Menus
         private void OnValueChanged()
         {
             InvokeOnChange(Value);
+            UpdateValueText();
+        }
+
+        private void UpdateValueText()
+        {
             if (_valueText != null)
             {
-                _valueText.text = Value.ToString();
+                _valueText.text = string.Format("{0:0.#}", Value);
             }
         }
 
--- a/OWML.ModHelper.Menus/ModSliderInput.cs
+++ b/OWML.ModHelper.Menus/ModSliderInput.cs
@@ -47,17 +47,16 @@ namespace OWML.ModHelper.Menus
         private Text GetValueText()
         {
             var slider = _element.GetComponentInChildren<Slider>();
-            if (slider != null)
-            {
-                return slider.GetComponentInChildren<Text>();
-            }
-            return null;
+            return slider?.GetComponentInChildren<Text>();
         }
 
         private void OnValueChanged()
         {
             InvokeOnChange(Value);
-            _valueText.text = Value.ToString();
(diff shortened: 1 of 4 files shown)


OWML.ModHelper.Menus/ModSliderInput.cs
```diff
@@ -1,27 +1,31 @@
...
             _element = element;
+            _valueText = GetValueText();
             _uIStyleApplier = element.GetComponent<UIStyleApplier>();
-            element.OnValueChanged += () => InvokeOnChange(Value);
+            element.OnValueChanged += OnValueChanged;
         }
 
-        public override float Value
-        {
+        public override float Value {
```
amazingalek, line 28: JS style 😋

OWML.ModHelper.Menus/ModSliderInput.cs
```diff
@@ -1,27 +1,31 @@
...
 using UnityEngine;
+using UnityEngine.UI;
 
 namespace OWML.ModHelper.Menus
 {
     public class ModSliderInput : ModInput<float>, IModSliderInput
     {
         public float Min { get; set; }
         public float Max { get; set; }
+        public bool HasValueText => _valueText != null;
```
amazingalek, line 12: Is this needed?

amazingalek:
Awesome! What does it look like in game?
Probably it's best to always show the value 😁

OWML.ModHelper.Menus/ModSliderInput.cs
Raicuparta: whaat, why didn't VS auto fix this 😠

OWML.ModHelper.Menus/ModSliderInput.cs
Raicuparta: Check this line: https://github.com/amazingalek/owml/pull/224/files#diff-a4fa79456d54d033ef6fe8039ce95a34R97

I use this so that the template slider element is one that has a value being displayed. For instance, FOV slider would be a good template (because it shows the value), but Gamma slider wouldn't.

FOV is the first slider in the graphics tab so getting the first slider would still work, but if they add a new slider or change the order in an update it could break.

Raicuparta:
> Awesome! What does it look like in game?
> Probably it's best to always show the value 😁

Looks just like the sliders with value that the game already has (like FOV slider). Didn't test with float values yet though, need to see. Think I'll need to limit the decimal points.

PR #225, opened 2020-07-23
moved RequireVR flag from config to manifest

TAImatem:
also removed that setting from mod's menu

### DIFF
--- a/OWML.Launcher/App.cs
+++ b/OWML.Launcher/App.cs
@@ -138,7 +138,7 @@ namespace OWML.Launcher
 
         private bool HasVrMod(IList<IModData> mods)
         {
-            var vrMod = mods.FirstOrDefault(x => x.Config.RequireVR && x.Config.Enabled);
+            var vrMod = mods.FirstOrDefault(x => x.Manifest.RequireVR && x.Config.Enabled);
             var hasVrMod = vrMod != null;
             _writer.WriteLine(hasVrMod ? $"{vrMod.Manifest.UniqueName} requires VR." : "No mods require VR.");
             return hasVrMod;
--- a/OWML.Launcher/App.cs
+++ b/OWML.Launcher/App.cs
@@ -138,7 +138,7 @@ namespace OWML.Launcher
 
         private bool HasVrMod(IList<IModData> mods)
         {
-            var vrMod = mods.FirstOrDefault(x => x.Manifest.RequireVR && x.Config.Enabled);
+            var vrMod = mods.FirstOrDefault(x => x.RequireVR && x.Config.Enabled);
             var hasVrMod = vrMod != null;
             _writer.WriteLine(hasVrMod ? $"{vrMod.Manifest.UniqueName} requires VR." : "No mods require VR.");
             return hasVrMod;
--- a/OWML.ModHelper/ModManifest.cs
+++ b/OWML.ModHelper/ModManifest.cs
@@ -37,7 +37,7 @@ namespace OWML.ModHelper
         public Dictionary<string, string> AppIds { get; private set; }
 
         [JsonProperty("requireVR")]
-        public bool RequireVR { get; set; }//set is to be made private once transition is over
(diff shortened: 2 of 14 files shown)


OWML.ModHelper/ModManifest.cs
```diff
@@ -36,6 +36,9 @@ public class ModManifest : IModManifest
         [JsonProperty("appIds")]
         public Dictionary<string, string> AppIds { get; private set; }
 
+        [JsonProperty("requireVR")]
+        public bool RequireVR { get; set; }//set is to be made private once transition is over
```
Raicuparta, line 40: Don't leave comments with the expectation of something being done about it in the future, because if it never happens then the comment just causes confusion. Make a GH issue (if it doesn't exist already) explaining the backwards-incompatible changes that will need to be made.

OWML.ModLoader/ModFinder.cs
```diff
@@ -42,6 +42,12 @@ private IModData GetModData(IModManifest manifest)
             var storage = new ModStorage(manifest);
             var config = storage.Load<ModConfig>(Constants.ModConfigFileName);
             var defaultConfig = storage.Load<ModConfig>(Constants.ModDefaultConfigFileName);
+            if (!manifest.RequireVR)//assume it wasn't present in manifest but present in config, to be removed
+            {
+                manifest.RequireVR = (config != null && config.RequireVR)
+                    || (config == null && defaultConfig != null && defaultConfig.RequireVR);
+                storage.Save(manifest, Constants.ModManifestFileName);
```
Raicuparta, line 49: Don't think OWML should ever edit a manifest file. Won't this work the same without this line anyway? Would just always set the internal manifest.RequireVR to whatever is in config, on every startup, but not save it to the file.

TAImatem: Hmmm... I guess you're correct. It's only intended to persist until 0.5.2 anyway

OWML.ModHelper/ModManifest.cs
TAImatem: it's expected to be removed in next update though

TAImatem: adding issue sounds reasonable though

Raicuparta: I know it's expected, but it's bad practice to try trust future predictions 😅

TAImatem: made a little list of RequireVR-related things to be removed later on

TAImatem:
I wonder If I should move (or remove altogether) requireVR flags from default-config to manifest in example mods :thinking:
PS I did test carry-over, works correctly

OWML.ModLoader/ModFinder.cs
```diff
@@ -42,6 +42,11 @@ private IModData GetModData(IModManifest manifest)
             var storage = new ModStorage(manifest);
             var config = storage.Load<ModConfig>(Constants.ModConfigFileName);
             var defaultConfig = storage.Load<ModConfig>(Constants.ModDefaultConfigFileName);
+            if (!manifest.RequireVR)
+            {
+                manifest.RequireVR = (config != null && config.RequireVR)
```
amazingalek, line 47: instead of changing the manifest here... (see other comment)

OWML.Launcher/App.cs
```diff
@@ -138,7 +138,7 @@ private void OnOutput(string s)
 
         private bool HasVrMod(IList<IModData> mods)
         {
-            var vrMod = mods.FirstOrDefault(x => x.Config.RequireVR && x.Config.Enabled);
+            var vrMod = mods.FirstOrDefault(x => x.Manifest.RequireVR && x.Config.Enabled);
```
amazingalek, line 141: ...how about also checking config here? (see other comment)

TAImatem: Maybe it's better to add RequireVR to ModData then?

PR #239, opened 2020-07-25
new logging project

amazingalek:


TAImatem:
Sounds like it *might* break something 🤔

TAImatem:
or is it 1.0 thing?

amazingalek:
@TAImatem I'm not sure how breaky this is, will test shortly.

TAImatem:
Although it might be fine since mods normally interact with interface, so technically if you somehow keep it the same in ModHelper it might work

PR #240, opened 2020-07-26
Made some fixes to Listener

TAImatem:


OWML.Launcher/SocketListener.cs
```diff
@@ -11,7 +11,8 @@ namespace OWML.Launcher
 {
     public class SocketListener
     {
-        private const int BufferSize = 1024;
+        private const string Separator = "\n--------------------------------";
+        private const int BufferSize = 1<<18;
```
misternebula, line 15: Either add a code comment here or write this out plainly - I can see someone (or us) having to look up what this is equal to

OWML.Launcher/SocketListener.cs
```diff
@@ -73,23 +78,40 @@ private void ListenToSocket()
...
 
         private void ProcessMessage(byte[] bytes, int count)
         {
-            var json = Encoding.UTF8.GetString(bytes, 0, count);
-
-            var data = JsonConvert.DeserializeObject<SocketMessage>(json);
-
-            if (data.Type == MessageType.Quit)
+            var message = Encoding.UTF8.GetString(bytes, 0, count);
+            var jsons = message.Split('\n');
```
misternebula, line 89: What if a message inside a json contains a \n

TAImatem: it shouldn't

TAImatem: inside messages themselves they are written as "\\n" which != '\n', and jsons are one-liners

TAImatem: so, unless you add newlines into json yourself it will work fine 😝

TAImatem: and I considered secret combination to be a bit iffy

amazingalek: > inside messages themselves they are written as "\\n"

Always? Whether log came from OWML or Unity?

OWML.Launcher/SocketListener.cs
```diff
@@ -73,23 +78,40 @@ private void ListenToSocket()
...
+            foreach (var json in jsons)
             {
-                Environment.Exit(0);
-            }
+                if (json == "")
```
amazingalek, line 92: `string.IsNullOrEmpty(json)`, maybe?

OWML.Launcher/SocketListener.cs
TAImatem: yes, it does that when converting to json string (and yes I tested it)

OWML.Launcher/SocketListener.cs
```diff
@@ -11,7 +11,8 @@ namespace OWML.Launcher
 {
     public class SocketListener
     {
-        private const int BufferSize = 1024;
+        private const string Separator = "\n--------------------------------";
+        private const int BufferSize = 262144;
```
amazingalek, line 15: Sure this won't clog up *something*?

OWML.Launcher/SocketListener.cs
amazingalek: If you say so ;)

TAImatem: as it works with both OWML message (one from Owo about changed settings) and Unity message (when testing with stack overflow trace)

OWML.Launcher/SocketListener.cs
TAImatem: worked fine with stack overflow trace (~90-100k symbols)

OWML.Launcher/SocketListener.cs
TAImatem: OK

TAImatem: though I'm gonna use IsNullOrWhiteSpace instead 😉

amazingalek:
Looks good but I'll let @misternebula have the final word ;)

PR #247, opened 2020-07-26
integrated taskdelayer into events

amazingalek:


TAImatem:
Why in ModHelper though?

OWML.Common/IModUnityEvents.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+
+namespace OWML.Common
+{
+    public interface IModUnityEvents
+    {
+        void FireOnNextUpdate(Action action);
```
amazingalek, line 7: Can add `event Action OnUpdate` and other Unity stuff here when/if needed later.

TAImatem: Might as well do it now I suppose

PR #257, opened 2020-07-26
copy config method

amazingalek:


OWML.ModLoader/ModData.cs
```diff
@@ -22,14 +22,14 @@ public class ModData : IModData
                                  Config != null && Config.RequireVR ||
                                  Config == null && DefaultConfig != null && DefaultConfig.RequireVR;
 
-        private IModConfig _configSnapshot;
+        private readonly IModConfig _configSnapshot;
```
TAImatem, line 25: won't really work if I would use Config.Copy in UpdatedSnapshot

amazingalek: ok, so like this? (pushed)

TAImatem:
Also add ?s just in case :wink:

TAImatem:
Did it myself 🤷‍♂️ https://github.com/amazingalek/owml/pull/262

amazingalek:
@TAImatem wanna merge this?

PR #262, opened 2020-07-27
made it handle null configs better

TAImatem:
PS yes, DefaultConfig can be null even after FixConfigs


TAImatem:
I hope it's safe to assume that FixConfigs will be called before RequireReload is requested

TAImatem:
Feel free to merge it yourself btw

PR #264, opened 2020-07-27
plsreload one liners

amazingalek:


OWML.ModLoader/ModData.cs
```diff
@@ -34,30 +34,12 @@ public ModData(IModManifest manifest, IModConfig config, IModConfig defaultConfi
...
-            }
-            else if (Config != null)
-            {
-                Config.Enabled = true; //to preserve other settings
-            }
-            else
-            {
-                Config = new ModConfig();
-            }
+            Config = DefaultConfig != null ? DefaultConfig.Copy() : new ModConfig();
```
TAImatem, line 42: WRONG 😝

TAImatem: you forgot that it would delete all other settings

amazingalek: that's the name of the method: ResetConfigToDefaults ? :)
if default config doesn't exist, resetting to new/empty config would give defaults that makes sense, at least to me.

TAImatem: problem is if default-config doesn't exist and config has some non-standard settings it would make stuff super annoying

TAImatem: and currently having default-config is not enforced 🤷‍♂️

TAImatem: otherwise we might as well do FixConfigs differently to actually kinda enforce having default-config

TAImatem: that is do
`if (DefaultConfig == null) DefaultConfig = new ModConfig();` at the start

PR #280, opened 2020-07-29
game version check

amazingalek:


OWML.Launcher/App.cs
```diff
@@ -59,6 +63,20 @@ public void Run(string[] args)
             Console.ReadLine();
         }
 
+        private void CheckVersion()
+        {
+            var owVersion = _gameVersion.GetVersion();
+            var supportedOwVersion = _owmlManifest.OWVersion;
+            if (owVersion.StartsWith(supportedOwVersion))
```
misternebula, line 70: hm. if there was a change that broke OWML from eg. 1.0.7 -> 1.0.7.2 (and the manifest says 1.0.7) this wouldn't hold up...

amazingalek: dunno what to do about it tho 🤷

OWML.Launcher/GameVersion.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        public GameVersion(IOwmlConfig config)
+        {
+            _config = config;
+        }
+
+        public string GetVersion()
+        {
+            var globalgamemanagers = File.ReadAllText($"{_config.DataPath}/globalgamemanagers");
+            return new Regex(VersionPattern).Match(globalgamemanagers).Value;
```
amazingalek, line 21: this might be a bit fragile - for now we only get one hit :P

PR #282, opened 2020-07-29
disabling mod menus if not v1.0.7

amazingalek:
This simply doesn't initialize mod menus unless game version starts with 1.0.7. 
OWML works otherwise fine, except: mods that try to do menu stuff will produce errors, but it doesn't seem to crash the game.
It's not a permanent fix but better than what we have now.


amazingalek:
Maybe we should just quit instead 🤷 
In any case, we are able to do version check in-game, as long as we do it before menus are initialized.

TAImatem:
I don't really think using in-code version is a good idea TBH

PR #283, opened 2020-07-29
debug mode fix

amazingalek:
There's a bug in EnableDebugMode, caused by an earlier cleanup i did.
To reproduce:

1. Make sure EnableDebugMode is enabled
2. Open config menu for EnableDebugMode
3. Edit a combination and leave menus
4. You get an error

### DIFF
--- a/OWML.SampleMods/OWML.EnableDebugMode/EnableDebugMode.cs
+++ b/OWML.SampleMods/OWML.EnableDebugMode/EnableDebugMode.cs
@@ -15,9 +15,9 @@ namespace OWML.EnableDebugMode
 
         public override void Configure(IModConfig config)
         {
-            foreach (var key in _inputs.Keys)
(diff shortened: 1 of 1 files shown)


TAImatem:
Ah, I suppose it's because I forgot to call `Clear`

TAImatem:
Also, I only just realized there might be a problem with me not filtering out combination from other strings (in form of extra errors).
Maybe we should add IsType method to config?

amazingalek:
> Ah, I suppose it's because I forgot to call `Clear`

You kinda did, I removed it in a cleanup 😥

amazingalek:
> Also, I only just realized there might be a problem with me not filtering out combination from other strings (in form of extra errors).
> Maybe we should add IsType method to config?

Maybe and maybe. We'll deal with it when/if we need to ;-)

PR #289, opened 2020-07-30
My take on version checking

TAImatem:
When below minimal version it will without any question (because it's obviously already tested and found to be incompatible)
When above maximal version it will ask whether or not it should continue loading. If player chooses to continue it will overwrite maximalVersion entry in manifest with current version to make it show up only once.
I chose to implement the latter this way to make time between new OW patch/release and new OWML patch/release a bit more comfortable

### DIFF
--- a/OWML.Launcher/App.cs
+++ b/OWML.Launcher/App.cs
@@ -74,29 +74,29 @@ namespace OWML.Launcher
 
         private void CheckGameVersion()
         {
-            var versionReader = new GameVersionReader(_writer, new BinaryPatcher(_owmlConfig, _writer));
+            var versionReader = new GameVersionReader(new BinaryPatcher(_owmlConfig, _writer));
             var gameVersionString = versionReader.GetGameVersion();
             _writer.WriteLine($"Game version: {gameVersionString}", MessageType.Info);
-            if (!Version.TryParse(gameVersionString, out var gameVersion))
+            var isValidFormat = Version.TryParse(gameVersionString, out var gameVersion);
+            var minVersion = new Version(_owmlManifest.MinGameVersion);
+            var maxVersion = new Version(_owmlManifest.MaxGameVersion);
+            if (!isValidFormat)
             {
                 _writer.WriteLine("Warning - non-standard game version formatting found", MessageType.Warning);
+            }
+            if (!isValidFormat || gameVersion > maxVersion)
+            {
                 PotentiallyUnsupported();
                 return;
             }
-            var minVersion = new Version(_owmlManifest.MinGameVersion);
-            var maxVersion = new Version(_owmlManifest.MaxGameVersion);
             if (gameVersion < minVersion)
             {
                 _writer.WriteLine("Unsupported game version found", MessageType.Error);
-                AnyKey();
-            }
-            if (gameVersion > maxVersion)
-            {
-                PotentiallyUnsupported();
+                AnyKeyExitConsole();
             }
         }
 
-        private void AnyKey()
+        private void AnyKeyExitConsole()
         {
             _writer.WriteLine("Press any key to exit...", MessageType.Info);
             Console.ReadKey();
--- a/OWML.Launcher/App.cs
+++ b/OWML.Launcher/App.cs
@@ -91,33 +91,43 @@ namespace OWML.Launcher
                 var maxNumber = Convert.ToInt32(splitMinVersion[i]);
                 if (gameNumber < minNumber)
                 {
-                    _writer.WriteLine("Unsupported game version found", MessageType.Error);
-                    _writer.WriteLine("Press any key to exit...", MessageType.Info);
-                    Console.ReadKey();
-                    ExitConsole();
+                    ThrowBelowError();
                 }
                 if (gameNumber > maxNumber)
                 {
-                    _writer.WriteLine("Potentially unsupported game version found", MessageType.Warning);
-                    ConsoleKey response;
-                    do
-                    {
-                        _writer.WriteLine("Continue loading? [y/n]", MessageType.Info);
-                        response = Console.ReadKey(false).Key;
-                        if (response != ConsoleKey.Enter)
-                        {
-                            Console.WriteLine();
-                        }
-                    } while (response != ConsoleKey.Y && response != ConsoleKey.N);
-                    if (response == ConsoleKey.N)
-                    {
-                        ExitConsole();
-                    }
-                    _owmlManifest.MaximalGameVersion = gameVersion;
-                    JsonHelper.SaveJsonObject(Constants.OwmlManifestFileName, _owmlManifest);
+                    ThrowAboveWarning(gameVersion);
                     return;
                 }
             }
+        }
+
+        private void ThrowBelowError()
+        {
+            _writer.WriteLine("Unsupported game version found", MessageType.Error);
+            _writer.WriteLine("Press any key to exit...", MessageType.Info);
+            Console.ReadKey();
+            ExitConsole();
+        }
+
+        private void ThrowAboveWarning(string currentVersion)
+        {
+            _writer.WriteLine("Potentially unsupported game version found", MessageType.Warning);
+            ConsoleKey response;
+            do
+            {
+                _writer.WriteLine("Continue loading? [y/n]", MessageType.Info);
+                response = Console.ReadKey(false).Key;
+                if (response != ConsoleKey.Enter)
+                {
+                    Console.WriteLine();
+                }
+            } while (response != ConsoleKey.Y && response != ConsoleKey.N);
+            if (response == ConsoleKey.N)
+            {
+                ExitConsole();
+            }
+            _owmlManifest.MaximalGameVersion = currentVersion;
+            JsonHelper.SaveJsonObject(Constants.OwmlManifestFileName, _owmlManifest);
         }
 
         private void CopyGameFiles()
--- a/OWML.Launcher/App.cs
+++ b/OWML.Launcher/App.cs
@@ -75,35 +75,50 @@ namespace OWML.Launcher
         private void CheckGameVersion()
         {
             var versionReader = new GameVersionReader(_writer, new BinaryPatcher(_owmlConfig, _writer));
-            var gameVersion = versionReader.GetGameVersion();
-            _writer.WriteLine($"Game version: {gameVersion}", MessageType.Info);
-            var splitGameVersion = gameVersion.Split('.');
-            var splitMinVersion = _owmlManifest.MinimalGameVersion.Split('.');
-            var splitMaxVersion = _owmlManifest.MaximalGameVersion.Split('.');
-            var count = Math.Min(splitGameVersion.Length, Math.Min(splitMaxVersion.Length, splitMinVersion.Length));
-            splitGameVersion[count - 1] = string.Join("", splitGameVersion.Skip(count - 1));
-            splitMinVersion[count - 1] = string.Join("", splitMinVersion.Skip(count - 1));
-            splitMaxVersion[count - 1] = string.Join("", splitMaxVersion.Skip(count - 1));
-            for (var i = 0; i < count; i++)
+            var gameVersionString = versionReader.GetGameVersion();
+            _writer.WriteLine($"Game version: {gameVersionString}", MessageType.Info);
+            try
             {
-                var gameNumber = Convert.ToInt32(splitGameVersion[i]);
-                var minNumber = Convert.ToInt32(splitMinVersion[i]);
-                var maxNumber = Convert.ToInt32(splitMaxVersion[i]);
-                if (gameNumber < minNumber)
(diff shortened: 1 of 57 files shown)


OWML.Launcher/App.cs
```diff
@@ -70,6 +72,54 @@ private void LocateGamePath()
...
+            var versionReader = new GameVersionReader(_writer, new BinaryPatcher(_owmlConfig, _writer));
+            var gameVersion = versionReader.GetGameVersion();
+            _writer.WriteLine($"Game version: {gameVersion}", MessageType.Info);
+            var splitGameVersion = gameVersion.Split('.');
+            var splitMinVersion = _owmlManifest.MinimalGameVersion.Split('.');
+            var splitMaxVersion = _owmlManifest.MaximalGameVersion.Split('.');
+            var count = Math.Min(splitGameVersion.Length, Math.Min(splitMaxVersion.Length, splitMinVersion.Length));
+            splitGameVersion[count-1] = string.Join("", splitGameVersion.Skip(count - 1));
+            splitMinVersion[count-1] = string.Join("", splitMinVersion.Skip(count - 1));
+            splitMaxVersion[count-1] = string.Join("", splitMaxVersion.Skip(count - 1));
```
TAImatem, line 86: not really sure about that part TBH. It feels like it might backfire in some way

OWML.Launcher/App.cs
```diff
@@ -70,6 +72,64 @@ private void LocateGamePath()
             }
         }
 
+        private void CheckGameVersion()
```
amazingalek, line 75: This is too low level for App.cs, can you move it to another/new class?

TAImatem: well, I can carry it to versionReader

amazingalek: C# has version comparison functionality built-in: https://www.c-sharpcorner.com/blogs/comparing-version-number-in-c-sharp
Use that and your code will be simple enough to stay where it is.

OWML.Launcher/App.cs
```diff
@@ -70,6 +72,64 @@ private void LocateGamePath()
...
+        {
+            _writer.WriteLine("Unsupported game version found", MessageType.Error);
+            _writer.WriteLine("Press any key to exit...", MessageType.Info);
+            Console.ReadKey();
+            ExitConsole();
+        }
+
+        private void ThrowAboveWarning(string currentVersion)
+        {
+            _writer.WriteLine("Potentially unsupported game version found", MessageType.Warning);
```
amazingalek, line 114: Since it's *potentially* unsupported, I don't think we should bother user with a prompt. I say we print a warning and continue. Keep it simple :)

OWML.Launcher/App.cs
```diff
@@ -70,6 +72,64 @@ private void LocateGamePath()
...
+                {
+                    Console.WriteLine();
+                }
+            } while (response != ConsoleKey.Y && response != ConsoleKey.N);
+            if (response == ConsoleKey.N)
+            {
+                ExitConsole();
+            }
+            _owmlManifest.MaximalGameVersion = currentVersion;
+            JsonHelper.SaveJsonObject(Constants.OwmlManifestFileName, _owmlManifest);
```
amazingalek, line 130: Why change the manifest here? We don't trust users to know whats supported or not?

Raicuparta: Agreed, I don't think we should ever edit manifest files.

Raicuparta: If something like this is needed, some other file should be used to persist the data (like the config file or something else)

OWML.Patcher/GameVersionReader.cs
```diff
@@ -0,0 +1,36 @@
...
+            _binaryPatcher = binaryPatcher;
+        }
+
+        public string GetGameVersion()
+        {
+            var sectorBytes = _binaryPatcher.GetSectorBytes(_binaryPatcher.ReadFileBytes(), PlayerSettingsSector);
+            for (int i = 4; i < sectorBytes.Length - 2; i++)
+            {
+                if (sectorBytes[i - 4] == 0 && sectorBytes[i - 3] == 0 && sectorBytes[i - 2] == 0 &&
+                    sectorBytes[i] == '.' && sectorBytes[i + 2] == '.')
```
amazingalek, line 27: Are we looking for [0, 0, 0, ?, '.', ?, '.']? Why, what's special about it? Seems fragile...

OWML.Launcher/App.cs
amazingalek: I guess it's to only prompt the user once. 
I think we should not bother the user at all, see above comment.

OWML.ModHelper/ModManifest.cs
```diff
@@ -41,5 +41,11 @@ public class ModManifest : IModManifest
 
         [JsonIgnore]
         public string AssemblyPath => ModFolderPath + Filename;
+
+        [JsonProperty("minimalGameVersion")]
+        public string MinimalGameVersion { get; private set; } = "";
+
+        [JsonProperty("maximalGameVersion")]
+        public string MaximalGameVersion { get; set; } = "";
```
amazingalek, line 49: I prefer MinGameVersion and MaxGameVersion, but if you wanna write it out, Minimum and Maximum is better.

OWML.Patcher/GameVersionReader.cs
TAImatem: I don't think first and second numbers would become two-digit anytime soon, as well as version string getting longet than 255 chars :P

OWML.ModHelper/ModManifest.cs
TAImatem: I too feel that `-imal` is a bit excessive

OWML.Launcher/App.cs
TAImatem: I will just turn it into warning then

OWML.Launcher/App.cs
```diff
@@ -75,35 +75,50 @@ private void LocateGamePath()
...
                 }
-                if (gameNumber > maxNumber)
+                if (gameVersion > maxVersion)
                 {
-                    ThrowAboveWarning(gameVersion);
+                    ThrowAboveWarning(gameVersion.ToString());
                     return;
                 }
             }
+            catch (Exception ex)
```
amazingalek, line 95: only `new Version(...)` can throw this, so move your comparisons out of `try`

OWML.Launcher/App.cs
```diff
@@ -75,35 +75,50 @@ private void LocateGamePath()
...
+            {
+                _writer.WriteLine($"Error while trying to perfrom version chec:\n{ex}", MessageType.Error);
+                AnyKey();
+            }
+        }
+
+        private string NormalizeVersionString(string version)
```
amazingalek, line 102: Why, though? This seems fragile. If OW suddenly has five components to the version I'd rather know about it: `$"Error - game version could not be parsed: {version}."`

OWML.Launcher/App.cs
TAImatem: I would have to define Version outside of `try..catch` then

OWML.Patcher/GameVersionReader.cs
TAImatem: or should I just do "two dots with character in-between" check?

OWML.Launcher/App.cs
```diff
@@ -70,6 +72,42 @@ private void LocateGamePath()
...
+            var maxVersion = new Version(_owmlManifest.MaxGameVersion);
+            if (gameVersion < minVersion)
+            {
+                _writer.WriteLine("Unsupported game version found", MessageType.Error);
+                AnyKey();
+            }
+            if (gameVersion > maxVersion)
+            {
+                PotentiallyUnsupported();
+            }
```
amazingalek, line 96: You can combine the checks a bit:
            ```
var minVersion = new Version(_owmlManifest.MinGameVersion);
            var maxVersion = new Version(_owmlManifest.MaxGameVersion);
            var isValidGameVersion = Version.TryParse(gameVersionString, out var gameVersion);
            if (!isValidGameVersion)
            {
                _writer.WriteLine("Warning - non-standard game version formatting found", MessageType.Warning);
            }
            if (gameVersion < minVersion)
            {
                _writer.WriteLine("Unsupported game version found", MessageType.Error);
                AnyKey();
            }
            if (gameVersion > maxVersion || !isValidGameVersion)
            {
                PotentiallyUnsupported();
            }
```

File: OWML.Patcher/GameVersionReader.cs (hunk 5 above)
Comment (amazingalek) [reply OWML.Patcher/GameVersionReader.cs:27]: Dunno but this is hard to read.
Can the whole thing be converted into a string, and then you look for the substring instead? That would be so much easier.

File: OWML.Patcher/GameVersionReader.cs [hunk 10]
```diff
@@ -0,0 +1,36 @@
...
+    public class GameVersionReader
+    {
+        private readonly IModConsole _writer;
+        private readonly BinaryPatcher _binaryPatcher;
+
+        private const int PlayerSettingsSector = 0;
+
+        public GameVersionReader(IModConsole writer, BinaryPatcher binaryPatcher)
+        {
+            _writer = writer;
```
amazingalek, line 17: not used

OWML.Launcher/App.cs
```diff
@@ -70,6 +72,42 @@ private void LocateGamePath()
...
+        }
+
+        private void AnyKey()
```
amazingalek, line 99: `AnyKeyAndQuit()` or similar

OWML.Patcher/GameVersionReader.cs
TAImatem: simple substring won't work

TAImatem: regex is nice though

OWML.Launcher/App.cs
TAImatem: done

OWML.Patcher/GameVersionReader.cs
TAImatem: done

OWML.Launcher/App.cs
TAImatem: `AnyKeyExitConsole()`

OWML.Patcher/GameVersionReader.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.Text.RegularExpressions;
+
+namespace OWML.Patcher
+{
+    public class GameVersionReader
+    {
+        private readonly BinaryPatcher _binaryPatcher;
+
+        private const int PlayerSettingsSector = 0;
+        private const string VersionPattern = @"\d+(\.\d+)+";//can handle 1-infinite dots
```
amazingalek, line 11: How many results does this give? There are dates tn the file that would give wrong results, no?

OWML.Patcher/GameVersionReader.cs
```diff
@@ -0,0 +1,26 @@
...
+        {
+            _binaryPatcher = binaryPatcher;
+        }
+
+        public string GetGameVersion()
+        {
+            var sectorBytes = _binaryPatcher.GetSectorBytes(_binaryPatcher.ReadFileBytes(), PlayerSettingsSector);
+            var sectorString = Encoding.ASCII.GetString(sectorBytes);
+            var match = Regex.Match(sectorString, VersionPattern);
+            return match.Success? match.Value: "not found";
```
amazingalek, line 23: formatting

OWML.Patcher/GameVersionReader.cs
TAImatem: Not in that sector

OWML.Patcher/GameVersionReader.cs
TAImatem: ?

TAImatem: ah, got it

OWML.Patcher/GameVersionReader.cs
TAImatem: For reference this sector is 0x1000-0x1257 (LeftStick_X_PC)

OWML.Patcher/GameVersionReader.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        private const int PlayerSettingsSector = 0;
+        private const string VersionPattern = @"\d+(\.\d+){3,}";//can handle 3-infinite dots
+
+        public GameVersionReader(BinaryPatcher binaryPatcher)
+        {
+            _binaryPatcher = binaryPatcher;
+        }
+
+        public string GetGameVersion()
```
amazingalek, line 18: since you're reading *all* bytes, whats the real point of using the binart patcher? this looks so similar to #280

TAImatem: I read only one sector

PR #291, opened 2020-07-31
InputHandler now responds to changes in OWML config immediately

TAImatem:
Who needs HarmonyHelper anyway 😝


TAImatem:
Probably should use `First` instead though

amazingalek:
Not sure about this. Seems hacky, and also no one uses the related feature anyway...

TAImatem:
You mean input blocking?

PR #292, opened 2020-07-31
re-remove game path

TAImatem:
poor lonely toogle


amazingalek:
@Raicuparta you said something about this which made me not remove it the last time. Thoughts?

Raicuparta:
> @Raicuparta you said something about this which made me not remove it the last time. Thoughts?

What? I did? Must not have been my intention. Game path can be changed in Mod Manager now so don't see the need for it in game. Do you remember what I said?

Issue #256, opened 2020-07-26
Make config files have only values


Raicuparta:
Should probably work the same way for mod configs and owml config if possible.

This would fix many config problems that we currently have (for instance can't change the max value of a slider in a mod update). Would also simplify the code in a few places in OWML (less logic for deciding whether settings need to be reset after mod updates) and Mod Manager (mod manager would no longer need to keep copying stuff from default).

- Structure of config could stay in either default-config.json, or move to manifest.json (might reduce the chance of accidentally editing defaults file?);
- config.json would only need to have values that differ from defaults;
- When OWML reads a setting value, it will first check config.json, then default-config.json if needed;
- We never need to copy from default-config.json to config.json;
- Pressing "reset to defaults" simply deletes config.json;
- Mod updates can change the format of settings (min/max values, titles, etc) and user configs remain intact;
- If a mod update changes the type of a setting (like string to int) then it needs to be reset (already done);

Example:

```
// default-config.json
{
  "enabled": true,
  "settings": {
    "enableMusic": true,
    "power": {
      "type": "slider",
      "value": 10,
      "min": 0,
      "max": 20
    },
    "decimals": {
      "type": "slider",
      "value": 10.54545,
      "min": 0.545,
      "max": 525.545845565
    }
  }
}
```

```
// config.json
{
  "enabled": false,
  "settings": {
    "power": 5,
    "decimals": 5.2
  }
}
```

PR #313, opened 2020-08-15
Use a single ModConfig that accesses user config and default config

Raicuparta:
Building towards #256

This doesn't have the value-only config.json thing yet. It just adds another ModConfig that merges both user config and default config together (so when you read from this single ModConfig, it checks both). It's still doing the whole thing of always copying default-config to user config, which is redundant. Will be simplified a bit once the value-only config is done.


OWML.Common/IModData.cs
```diff
@@ -8,9 +8,9 @@ public interface IModData
         bool Enabled { get; }
         bool RequireVR { get; }
         bool RequireReload { get; }
+        IModConfig UserConfig { get; }
 
         void UpdateSnapshot();
         void ResetConfigToDefaults();
-        bool FixConfigs();
```
Raicuparta, line 14: I could keep this for backwards compatibility, but it wouldn't do anything. And I doubt it would break any mod anyway.

OWML.ModLoader/ModData.cs
```diff
@@ -11,134 +11,34 @@ public class ModData : IModData
     {
         public IModManifest Manifest { get; }
         public IModConfig Config { get; private set; }
+        [Obsolete("Maybe this should be obsolete?")]
         public IModConfig DefaultConfig { get; private set; }
```
Raicuparta, line 15: As far as I can tell this isn't used anywhere else in OWML, so it could be removed. Could mark as obsolete, but there isn't really an alternative.

OWML.ModHelper/ModMergedConfig.cs
```diff
@@ -0,0 +1,162 @@
...
+        }
+
+        private bool IsSettingSameType(object settingValue1, object settingValue2)
+        {
+            return settingValue1.GetType() == settingValue2.GetType() &&
+                   (!(settingValue1 is JObject obj1) || !(settingValue2 is JObject obj2) ||
+                    (string)obj1["type"] == (string)obj2["type"]);
+        }
+    }
+}
```
Raicuparta, line 162: All of this code is just moved straight from ModData.cs

OWML.ModHelper/ModMergedConfig.cs
```diff
@@ -0,0 +1,162 @@
...
+        public IModConfig Copy()
+        {
+            return new ModConfig
+            {
+                Enabled = Enabled,
+                Settings = new Dictionary<string, object>(Settings)
+            };
+        }
+
+        private void FixConfigs()
```
Raicuparta, line 58: All of the code starting with this method until the end of this class is just copied straight from ModData.cs, with minor adjustments to make it work here instead. This code would be greatly simplified after the next PR is done, which will make config.json have only values, and thus have a much simpler process for finding inconsistencies.

OWML.ModHelper/ModMergedConfig.cs
```diff
@@ -0,0 +1,162 @@
...
+            FixConfigs();
+        }
+
+        public T GetSettingsValue<T>(string key)
+        {
+            if (_userConfig.Settings.ContainsKey(key))
+            {
+                return _userConfig.GetSettingsValue<T>(key);
+            }
+            return _defaultConfig.GetSettingsValue<T>(key);
```
amazingalek, line 41: i *think* you can do 
`return `_userConfig.GetSettingsValue<T>(key) ?? _defaultConfig.GetSettingsValue<T>(key);`

OWML.ModLoader/ModData.cs
amazingalek: try removing it, if it causes issues we'll know when testing ;)

OWML.ModHelper/ModMergedConfig.cs
Raicuparta: I started off that way, but get settings value isn't prepared for missing keys.

OWML.ModLoader/ModData.cs
```diff
@@ -11,134 +11,34 @@ public class ModData : IModData
...
+
 
         private IModConfig _configSnapshot;
 
         public ModData(IModManifest manifest, IModConfig config, IModConfig defaultConfig)
         {
             Manifest = manifest;
-            Config = config;
-            DefaultConfig = defaultConfig ?? new ModConfig();
+            Config = new ModMergedConfig(config, defaultConfig, manifest);
```
amazingalek, line 28: Config -> MergedConfig?

OWML.ModHelper/ModMergedConfig.cs
amazingalek: Ah, then please prepare it for that ;)

amazingalek: Return default value of the relevant type, in that case

OWML.ModLoader/ModData.cs
Raicuparta: I was trying to keep it as Config to preserve backwards compatibility. Does that not matter here? I guess mods aren't supposed to access this directly.

This code could be a lot cleaner if ModMergedConfig didn't have to be of type IModConfig, so if I don't have to worry about backwards compatibility with Config I can clean this up a lot.

OWML.ModHelper/ModMergedConfig.cs
Raicuparta: Ah yeah, that was the problem. It returns the default value of the type, which I don't want. So I need to manually check the dictionary.

OWML.ModLoader/ModData.cs
Raicuparta: I mean ModMergedConfig would still implement IModConfig, but in ModData it would be declared as type ModMergedConfig, so that I can call ModMergedConfig specific methods, no need to do some of the workarounds I'm doing now. But only if backwards compatibility doesn't matter here.

amazingalek: from memory, whats importantnisnthat mod.Configure is called with IModConfig. that might be the only thing that matters. 
General tip: experiment, test with all mods. No errors -> it's fine ;-)

Raicuparta: Made the change and tested with all mods. Had to fix an error with missing configs, other than that seems to work fine :)

Raicuparta: I left it as "Config" since now there's no UserConfig. Should I change it to MergedConfig anyway?

PR #315, opened 2020-08-17
Value-only user config

### LINKED ISSUE #256: Make config files have only values

Raicuparta:
Working towards #256

Some stuff still missing from this perfect config plan:
* when user changes a setting, it should only add that single setting to config.json, instead of always adding everything;
* check for inconsistencies with values of sliders and selectors (need to be within possible values);

These will be done in a separate PR

### DIFF
--- a/OWML.Common/IModConfig.cs
+++ b/OWML.Common/IModConfig.cs
@@ -10,7 +10,5 @@ namespace OWML.Common
 
         T GetSettingsValue<T>(string key);
         void SetSettingsValue(string key, object value);
-
-        IModConfig Copy();
     }
 }
--- a/OWML.ModHelper/ModMergedConfig.cs
+++ b/OWML.ModHelper/ModMergedConfig.cs
@@ -15,7 +15,7 @@ namespace OWML.ModHelper
             set => _userConfig.Settings = value;
         }
 
-        private IModConfig _userConfig;
+        private readonly IModConfig _userConfig;
         private readonly IModConfig _defaultConfig;
         private readonly IModStorage _storage;
 
@@ -38,7 +38,7 @@ namespace OWML.ModHelper
 
         public void SetSettingsValue(string key, object value)
         {
-            _userConfig.SetSettingsValue(key, value);
+            _userConfig.SetSettingsValue(key, GetConvertedSelectorValue(key, value) ?? value);
         }
 
         public IModConfig Copy()
@@ -106,6 +106,21 @@ namespace OWML.ModHelper
                 return true;
             }
             return userValue.GetType() == defaultInnerValue.GetType();
+        }
+
+        private object GetConvertedSelectorValue(string key, object value)
+        {
+            if (!_defaultConfig.Settings.ContainsKey(key))
+            {
+                return null;
+            }
+            var defaultValue = _defaultConfig.Settings[key];
+            if (defaultValue is JObject defaultObjectValue && defaultObjectValue["type"].ToString() == "selector")
+            {
+                var innerValue = defaultObjectValue["value"];
+                return innerValue.Type == JTokenType.Integer ? int.Parse(value.ToString()) : value;
+            }
+            return null;
         }
     }
 }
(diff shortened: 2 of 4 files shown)


OWML.Common/IModConfig.cs
```diff
@@ -10,7 +10,5 @@ public interface IModConfig
 
         T GetSettingsValue<T>(string key);
         void SetSettingsValue(string key, object value);
-
-        IModConfig Copy();
```
Raicuparta, line 14: This was only being used to create the snapshot, which is now only needed in ModMergedConfig, not in defaultConfig or userConfig

OWML.ModHelper/ModMergedConfig.cs
```diff
@@ -60,120 +57,83 @@ public void SaveToStorage()
...
+            {
+                return null;
+            }
+            var defaultValue = _defaultConfig.Settings[key];
+            if (defaultValue is JObject defaultObjectValue && defaultObjectValue["type"].ToString() == "selector")
+            {
+                var innerValue = defaultObjectValue["value"];
+                return innerValue.Type == JTokenType.Integer ? int.Parse(value.ToString()) : value;
+            }
+            return null;
```
Raicuparta, line 136: This is the hackiest part of this change. It was already broken before, but in a way that didn't affect anything. Our selectors are created as extending `IModInput<string>`, which doesn't make much sense since selectors can have string or int values. I couldn't find a good way to fix this at the source, like making our selectors generic, or having two types of selector. So instead I'm parsing the string value back to int for number selectors. The type of selector is determined by the type of the default value.

OWML.ModHelper/ModMergedConfig.cs
```diff
@@ -60,120 +57,85 @@ public void SaveToStorage()
...
+        private bool IsNumber(object value)
         {
-            var missingSettings = defaultConfig.Settings.Where(s => !_userConfig.Settings.ContainsKey(s.Key)).ToList();
-            missingSettings.ForEach(setting => _userConfig.Settings.Add(setting.Key, setting.Value));
+            return value is int
+                || value is long
+                || value is float
+                || value is double
+                || value is decimal
+                || value is ulong;
```
Raicuparta, line 91: I feel like this could be much simpler, but I don't seem to understand how Json.Net decides to which type it will convert a JSON property. From my tests it seems like our numbers are always either long or float, but I have no idea if it will suddenly decide to use some other number type, so I used all the types that are listed in Json.Net's documentation.

A lot of this stuff could be heavily simplified if we had our own serializer, meaning we would always know exactly what types we're dealing with, and we wouldn't need to mess directly with `object`, `JObject`, etc, and convert them back and forth so much.

PR #344, opened 2021-10-01
Fix game version finding + remove VR patch

misternebula:


src/OWML.Launcher/App.cs
```diff
@@ -58,7 +55,7 @@ public void Run()
 
 			ShowModList(mods);
 
-			PatchGame(mods);
+			_owPatcher.PatchGame();
```
TAImatem, line 58: huh?

TAImatem:
"updated supported versions" that seems a bit optimistic :P
anyway, seems alright, but if it can wait I'd wait till tommorow

amazingalek (approved):

PR #351, opened 2021-10-06
ModSocket fails gracefully now

TAImatem:
self-explanatory, I guess


misternebula:
I would add an error message yeah. Other than that, looks good.

PR #362, opened 2021-10-12
2.0.0

misternebula:


amazingalek:
we should test the nuget package before merging

PR #397, opened 2021-10-29
Removed obsolete code

amazingalek:
This might break some mods. Should have done it in 2.0.0 but forgot 😢


src/OWML.Logging/ModLogger.cs
```diff
@@ -1,35 +0,0 @@
-﻿using System;
-using System.IO;
-using System.Linq;
-using OWML.Common;
-
-namespace OWML.Logging
-{
-	public class ModLogger : IModLogger
```
misternebula, line 8: so are we not logging to files anymore?

amazingalek:
Maybe later 🙃

PR #400, opened 2021-10-30
fix manifest versions 1.1.10 - 1.1.11

Raicuparta:


TAImatem (approved):
weird that both fields are hard requirments now, but that's definitely out of scope of this PR

PR #402, opened 2021-10-31
Option to launch through exe

misternebula:


src/OWML.Launcher/App.cs
```diff
@@ -166,22 +166,32 @@ private void StartGame()
 
 			try
 			{
-				if (_owmlConfig.GamePath.Contains("Epic"))
+				void StartGameViaExe()
+				{
+					_writer.WriteLine("Starting game via exe...");
+					_processHelper.Start(_owmlConfig.ExePath, _argumentHelper.Arguments);
```
amazingalek, line 172: maybe just move these lines to line 177 instead of calling this inline method?

amazingalek (approved):

src/OWML.Launcher/App.cs
misternebula: but this code is also called on line 193

PR #405, opened 2021-11-04
Updated ModMenu.cs

salomj:
Some minor refactoring to make it more obvious what is happening.

### DIFF
--- a/src/OWML.ModHelper.Menus/ModMenu.cs
+++ b/src/OWML.ModHelper.Menus/ModMenu.cs
@@ -237,12 +237,10 @@ namespace OWML.ModHelper.Menus
 			{
 				return numberInput.Value;
 			}
-			var separator = GetSeparator(key);
-			if (separator != null)
+			if (GetSeparator(key) == null)
 			{
-				return null;
+				Console.WriteLine($"Error - No input found with name {key}", MessageType.Error);
 			}
-			Console.WriteLine($"Error - No input found with name {key}", MessageType.Error);
 			return null;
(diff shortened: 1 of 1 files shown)


src/OWML.ModHelper.Menus/ModMenu.cs
```diff
@@ -237,12 +237,10 @@ public object GetInputValue(string key)
 			{
 				return numberInput.Value;
 			}
-			var separator = GetSeparator(key);
-			if (separator != null)
+			if (GetSeparator(key) == null)
```
TAImatem, line 240: Should Separator even be considered as input if null is returned regardless? Maybe just completely remove that check

src/OWML.ModHelper.Menus/ModMenu.cs
```diff
@@ -283,12 +281,10 @@ public void SetInputValue(string key, object value)
 				numberInput.Value = Convert.ToSingle(val);
 				return;
 			}
-			var separator = GetSeparator(key);
-			if (separator != null)
+			if (GetSeparator(key) == null)
```
TAImatem, line 284: Should Separator even be considered as input if null is returned regardless? Maybe just completely remove that check

src/OWML.ModHelper.Menus/ModMenu.cs
salomj: The issue is that it will create a log message without the check.

Raicuparta (approved):

PR #411, opened 2021-11-15
Make it easier for first time contributors

salomj:
Now just add `OuterWildsRootDirectory` property to your `*.csproj.user`.


amazingalek:
I think this won't be necessary after https://github.com/amazingalek/owml/pull/413 ?

salomj:
> I think this won't be necessary after #413 ?

Maybe not. I'll check after it is merged.

Raicuparta:
No longer necessary after #419

PR #443, opened 2022-06-06
2.4.0

misternebula:
- Fixed audioclips being the wrong length
- Added optional parameters to JsonHelper to use settings and disable backslash replacement
- Added new methods to fix NRE in ModInteraction. Obsolete-d old methods.


src/OWML.ModHelper.Interaction/ModInteraction.cs
```diff
@@ -65,12 +66,41 @@ public IList<IModBehaviour> GetDependencies(string uniqueName)
 			return _dependencyDict[uniqueName];
 		}
 
+		public IModBehaviour TryGetMod(string uniqueName)
+			=> _modList.FirstOrDefault(m => m.ModHelper.Manifest.UniqueName == uniqueName);
+
+		private object TryGetApi(string uniqueName)
+		{
+			var mod = TryGetMod(uniqueName);
+			return mod == default ? default : mod.Api;
```
amazingalek, line 75: i think this would work: `return mod?.Api;`
not important tho, it's fine as it is

src/OWML.ModHelper.Interaction/ModInteraction.cs
```diff
@@ -65,12 +66,41 @@ public IList<IModBehaviour> GetDependencies(string uniqueName)
...
+			{
+				return default;
+			}
+
+			return api switch
+			{
+				null => null,
+				TInterface inter => inter,
+				_ => _proxyFactory.CreateProxy<TInterface>(api, _manifest.UniqueName, uniqueName)
+			};
```
amazingalek, line 92: nice :D

amazingalek (approved):
👍

src/OWML.ModHelper.Interaction/ModInteraction.cs
Raicuparta: mod is a ModBehaviour, right? MonoBehaviours have the unity lifetime check thing, which doesn't work with optional chaining

PR #447, opened 2022-06-07
2.4.1

misternebula:
Add JSON changes into ModStorage method


src/OWML.Launcher/OWML.Manifest.json
```diff
@@ -2,7 +2,7 @@
   "author": "Alek",
   "name": "OWML",
   "uniqueName": "Alek.OWML",
-  "version": "2.4.0",
+  "version": "2.4.1",
```
Raicuparta, line 5: This has to be a minor bump again. Easy way to tng of this: if it were a patch bump, it would mean people a mod made with owml 2.4.1 would work with owns 2.4.0, which isn't the case for this update.

misternebula: done 👍

Raicuparta (approved):

PR #450, opened 2022-06-08
2.5.1

misternebula:


Raicuparta (approved):

src/OWML.Common/Interfaces/IModStorage.cs
```diff
@@ -1,10 +1,14 @@
 ﻿using Newtonsoft.Json;
+using System;
 
 namespace OWML.Common
 {
 	public interface IModStorage
 	{
-		T Load<T>(string filename, bool fixBackslashes = true, JsonSerializerSettings settings = null);
+		[Obsolete]
```
Raicuparta, line 8: Why obsolete? And if obsolete, why no message?

Raicuparta: If it's because fix slashes defaults to false, then better add a message about that

misternebula: removed obsolete attribute

PR #453, opened 2022-06-13
Fixed #452 (exceptions in mod settings break options for other mods)

### LINKED ISSUE #452: If one mod throws an exception in Configure all mod settings break
Seems that when you change the settings for one mod, it runs `Configure` on all mods. If a mod throws an exception (happens often on the main menu because many creators don't handle NRE stuff properly since they expect you to be in the main scene) then your settings don't get updated even if you're trying to change the value for a mod that has no issues.

Specifically, happens to me when changing New Horizons settings from the main menu with Third Person Camera installed. Camera mod throws an exception so the New Horizons `Configure` doesn't run.

xen-42:


src/OWML.ModHelper.Menus/ModConfigMenu.cs
```diff
@@ -1,10 +1,11 @@
 ﻿using OWML.Common;
-using System.Linq;
 using OWML.Common.Menus;
+using System;
+using System.Linq;
 
 namespace OWML.ModHelper.Menus
 {
-	public class ModConfigMenu : ModConfigMenuBase, IModConfigMenu
+    public class ModConfigMenu : ModConfigMenuBase, IModConfigMenu
```
Raicuparta, line 8: This one

Raicuparta (approved):

PR #470, opened 2022-08-21
Fix harmony dependency in nuget

MegaPiggy:
1.2.0.1 doesn't use HarmonyLib as a namespace

It should be HarmonyX instead of regular harmony
That is what HarmonyHelper uses after all


JohnCorby:
PLEASE do this. I am tired of having to manually include HarmonyX every time lol

PR #475, opened 2022-09-05
Dev -> master

misternebula:


xen-42 (approved):
Seems fine to me, although I question why we're able to change the settings for mods that are not enabled.

MegaPiggy:
Pls merge

PR #496, opened 2022-10-09
Fix all mods config resetting when clicking the default settings button in any menu

dgarroDC:
All mods config menus added their `OnReset` on the `OnClick` of the reset button when initializing the menu, meaning that clicking the button makes all mod configs reset, even in vanilla settings menus like Gameplay and Graphics. The solution is adding the `OnReset` only when the menu is activated, and removing it when the menu is deactivated.

The same applies to the "save and exit" (close) button, although this probably didn't cause any bugs.


JohnCorby (approved):
ye that makes sense. silly alek or rai or whoever made this feature

PR #524, opened 2023-02-08
Switch to .NET Framework 4.7, update HarmonyX

misternebula:


MegaPiggy:
Why switch from 4.8 to 4.7?

misternebula:
So it works on Mono - https://www.mono-project.com/docs/about-mono/compatibility/

JohnCorby:
it already works with mono, as tested by bunnie and idiot

PR #540, opened 2023-07-02
Only sort enabled mods by order

### LINKED ISSUE #541: Disabled mod with cyclic dependency breaks load order

xen-42:
Currently the mod loader can't support cyclic mod dependencies. This doesn't fix that. It however checks if disabled mods have cyclic dependencies even though they aren't going to be loaded. This PR has it ignore disabled mods when sorting.

Else you can have a mod with cyclic dependency, disable it, and still have your load order broken on all other mods. Was happening with outdated Outsider Translation mod.

Fixes #541


xen-42:
Also should this be merged into dev or can we skip right into master

Bwc9876 (approved):
LGTM!

PR #554, opened 2023-08-17
2.9.6

misternebula:
APIs can now use custom enums. The enums need to be identical on both sides. That means the :
- Type name
- Amount of entries in the enum
- The underlying type (int, uint, ulong, etc.)
- Value of each entry

has to be the same.

### DIFF
--- a/src/OWML.ModHelper.Interaction/InterfaceProxyBuilder.cs
+++ b/src/OWML.ModHelper.Interaction/InterfaceProxyBuilder.cs
@@ -180,7 +180,7 @@ namespace OWML.ModHelper.Interaction
 
 			// Equivalent code :
 
-			// return instance.method(arg1, arg2, ...);
+			// return __Target.methodname(arg1, arg2, ...);
 		}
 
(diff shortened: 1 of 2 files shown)


src/OWML.ModHelper.Interaction/InterfaceProxyBuilder.cs
```diff
@@ -76,6 +76,13 @@ bool AreTypesMatching(Type a, Type b)
 					return false;
 				}
 
+				if (a.IsEnum && b.IsEnum)
+				{
+					// this doesn't check if the actual enum values are identical.... too bad!
```
JohnCorby, line 81: probably should do that

JohnCorby (changes requested):

src/OWML.ModHelper.Interaction/InterfaceProxyBuilder.cs
misternebula: done

JohnCorby (approved):

PR #588, opened 2024-07-04
Add Unpatching

Bwc9876:
Adds the ability for mods to specify an `unpatcher`.

This allows mods that use prepatchers to undo what those prepatchers did once disabled. Meaning users won't need to verify game files or undo changes manually.

## Added

- New manifest field: `unpatcher`
- New OWML config field: `prepatchersExecuted`
- New docs page on prepatching
- Manager now hides prepatcher warnings for mods with unpatchers on disable

### DIFF
--- a/.idea/.idea.OWML/.idea/.gitignore
+++ b/.idea/.idea.OWML/.idea/.gitignore
@@ -1,0 +1,13 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Rider ignored files
+/contentModel.xml
+/projectSettingsUpdater.xml
+/.idea.OWML.iml
+/modules.xml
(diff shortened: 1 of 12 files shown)


JohnCorby:
regarding docs, it would be nice to see if getexecutingassembly works for getting the mod path. then we could just be like "use appdomain blah blah for game path and getexecutingassembly for mod path"

MegaPiggy:
> regarding docs, it would be nice to see if getexecutingassembly works for getting the mod path. then we could just be like "use appdomain blah blah for game path and getexecutingassembly for mod path"

isn't one of the args the mod path?

Bwc9876:
kill

PR #597, opened 2024-09-26
2.14.0

MegaPiggy:
- Fixed issue where menus weren't able to be used with controller/keyboard.
- Added the ability to put negative numbers in the config menu's number fields.
  - Deprecated `OnInputPopupValidateChar` and added fixed version as `OnValidateChar`. Don't think anyone had used this event yet, but just in case.
- Improved look of some config menu items.
- Fixed "MOD OPTIONS" tab closing when clicked on.
- Made "MODS" selected by default, instead of "OWML".
- Fixed text/number entry popup forcing you back to the GAMEPLAY tab.
- Added `SetValue()` and `SetText()` to IOWMLTextEntryElement.
- Made `GameObjectHelper.CreateAndAdd` use the type name if no object name is provided.


src/OWML.ModHelper.Menus/CustomInputs/OWMLTextEntryElement.cs
```diff
@@ -23,9 +23,14 @@ public string GetInputText()
 			return _popup.GetInputText();
 		}
 
-		public void SetCurrentValue(string text)
```
Bwc9876, line 26: Do we know if any mods depend on this method? Would this break them?

src/OWML.Common/Interfaces/Menus/IOWMLPopupInputMenu.cs
```diff
@@ -4,7 +4,9 @@ namespace OWML.Common.Interfaces.Menus
 {
 	public interface IOWMLPopupInputMenu
 	{
-		public event PopupInputMenu.InputPopupValidateCharEvent OnInputPopupValidateChar;
+		public delegate bool InputPopupValidateCharEvent(string input, int charIndex, char addedChar);
```
Bwc9876, line 7: Would this break mods that use this expecting only a `char` to be passed?

src/OWML.ModHelper.Menus/CustomInputs/OWMLTextEntryElement.cs
misternebula: Dont think so, but changed back to be save.

src/OWML.Common/Interfaces/Menus/IOWMLPopupInputMenu.cs
MegaPiggy: yes but considering no mod has used it the event yet I think it is fine

misternebula (approved):

PR #602, opened 2024-12-11
Add Prepatching to Docs

Bwc9876:


JohnCorby (approved):

JohnCorby (changes requested):
link to qsb and nomaivr as samples. or maybe not those ones if you can think of better ones

JohnCorby (approved):

PR #604, opened 2024-12-20
Multipart Mod Proposal

Bwc9876:
Due to the advent of larger mods in development, we'll eventually need to deal with mods that are too large for 2 GB release asset limit. After talking on the Discord our current idea is a new field in the manifest that tells the manager to download additional ZIP files by name from the release of the mod (simple enough to just get the parent of the `downloadUrl` from the db and append the name to that). The manager would then extract these zip on top of the current mod's folder, effectively combining them all.

### DIFF
--- a/schemas/manifest_schema.json
+++ b/schemas/manifest_schema.json
@@ -129,6 +129,13 @@
                 ]
             }
         },
+        "additionalAssets": {
+            "type": "array",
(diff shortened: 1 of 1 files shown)


JohnCorby (approved):
could mention that the 2gb is per-file but i dont think thats necessary

xen-42 (approved):
Should probably document that the db tracks the first zip file in a release, to make it clear what an additional zip file is

PR #607, opened 2025-01-13
Updated file logging

misternebula:
- Added `LoadTime` to OWML config.
- Written log files are now based on the ISO 8601 standard (`OWML.Log.YYYY-MM-DDThh:mm:ss.txt`).
   - This applies to normal log files and Harmony log files.
- Added `latest.txt`. Contains the same data as the newest log file, but is overwritten on OWML load.


Bwc9876 (approved):
lgtm

PR #625, opened 2026-04-11
Make developing easier

MegaPiggy:
I was annoyed that
- I had to copy manually each time I built
- There was no line numbers

so here is a PR to fix that


JohnCorby (approved):
might be worth bumping LangVersion to latest? but doesnt matter

PR #626, opened 2026-04-11
Separator Dots and Label

MegaPiggy:
- Added a label to settings
- Added a "dots" field to separators. The default is true.
- Configs now always update with everything new in default config instead of only title and tooltip

Inspired by Ship Enhancements mod

### DIFF
--- a/schemas/config_schema.json
+++ b/schemas/config_schema.json
@@ -53,6 +53,7 @@
                         "type": {
                             "type": "string",
                             "enum": [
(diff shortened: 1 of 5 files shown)


JohnCorby (approved):
assuming megapiggy tested

JohnCorby (changes requested):
oops no forgot to update schema

JohnCorby (approved):

PR #629, opened 2026-04-26
2.16.1

misternebula:
- Fixed major bug where game would fail to initialize properly when using running on Linux/MacOS with Proton/Wine/Crossover.


Bwc9876 (approved):
Looks good to me!

PR #630, opened 2026-04-26
add mouse and fix keyboard path

MegaPiggy:
- Adds a method to set mouse inputs as initial bindings
- Adds a way to set the initial bindings to nothing (good for when you want to make a keyboard input but don't have something on the gamepad for it)
- Fixed a bug where the control path and groups weren't being set properly on the bindings.
- Added `FlipImages` to `RebindingHelper`. Used for horizontal axes.

- Also just converted all `.parent = ` to `SetParent` because the warnings annoyed me


xen-42 (approved):
lgtm and tested and it fixed the bug where the input rebinding didnt work

PR #638, opened 2026-09-04
2.16.3

misternebula:
- Fixed text entry inputs not being affected by the Default Settings button. (Thanks @milliero-se!)
- Adds everything to a `FontAndLanguageController` - should fix missing text when playing in non-latin languages. (#573)

### DIFF
--- a/src/OWML.ModHelper.Menus/NewMenuSystem/MenuUtilities.cs
+++ b/src/OWML.ModHelper.Menus/NewMenuSystem/MenuUtilities.cs
@@ -1,0 +1,28 @@
+﻿using System.Linq;
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace OWML.ModHelper.Menus.NewMenuSystem
+{
+	internal static class MenuUtilities
+	{
+		internal static void AddToLangController(Text textComponent)
+		{
+			var controllers = Resources.FindObjectsOfTypeAll<FontAndLanguageController>();
+
+			if (controllers.Length == 0)
+			{
+				return;
+			}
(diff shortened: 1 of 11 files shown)


src/OWML.ModHelper.Menus/NewMenuSystem/MenuUtilities.cs
```diff
@@ -0,0 +1,28 @@
...
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace OWML.ModHelper.Menus.NewMenuSystem
+{
+	internal static class MenuUtilities
+	{
+		internal static void AddToLangController(Text textComponent)
+		{
+			var controllers = Resources.FindObjectsOfTypeAll<FontAndLanguageController>();
```
JohnCorby, line 11: is this gonna be an issue perf wise? probably not because this only runs when its initially making the menus right?

JohnCorby (approved):

src/OWML.ModHelper.Menus/NewMenuSystem/MenuUtilities.cs
misternebula: yea only runs once

misternebula: it would be better if i created a single langcontroller in MenuManager, but i couldn't remember how to add a component in owml

milliero-se:
Wow, thanks for getting this merged so quickly! I appreciate the work y'all have put into this. :)

